PR #1, opened 2019-11-16
helm 3 crd support

jeffhollan:
resolves https://github.com/kedacore/keda/issues/406

This is a backwards compat way so both helm 2 and helm 3 will work with chart install


ahmelsayed:
Thank you Jeff! I could have sworn I tried that but it didn't work for some reason. I tried your change and it seems to work fine. I'll merge it and update the index.yaml

jeffhollan:
Ah thanks @ahmelsayed - I didn't realize we have the index.yaml here too.  Thanks - if you don't have time to get to it I can try to update later as well

jeffhollan:
ah jk looks like you already got it.  Thanks!

PR #2, opened 2019-11-18
add default log level for operator

zroubalik:
Adding default log level to deployment.yaml. So users can easily find the right place, in case they would need to change the log level.

Fixes https://github.com/kedacore/keda/issues/467


jeffhollan:
Would be nice to make this a template value so that if folks want they can set a parameter to change from `info` to `1`

ahmelsayed:
Fixed in #9

PR #7, opened 2019-11-28
updating crd to work with 1.0.0 image

jeffhollan:
related to https://github.com/kedacore/keda/issues/487 and https://github.com/kedacore/keda/issues/489


jeffhollan:
Gonna merge this in and manually test given there is a bug in the chart right now and right before long weekend.  Can always revert if I made a terrible mistake 😅 🤞

PR #23, opened 2020-02-17
logging levels for Operator and Metrics Server

zroubalik:
* enables us to modify log level for Metrics Server
* documents possible log levels for both Operator and Metrics server

Related issue: https://github.com/kedacore/keda/issues/618

### DIFF
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -36,7 +36,15 @@ aadPodIdentity: ""
 # will be mounted to the /grpccerts path on the Pod
 grpcTLSCertsSecret: ""
 
+## Logging level for KEDA Controller 
+# allowed values: 'debug', 'info', 'error', or an integer value greater than 0, specified as string
+# default value: info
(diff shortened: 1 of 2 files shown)


keda/values.yaml
```diff
@@ -36,8 +36,16 @@ aadPodIdentity: ""
 # will be mounted to the /grpccerts path on the Pod
 grpcTLSCertsSecret: ""
 
+## Logging level for KEDA Controller
```
tomkerkhove: Can we somehow align the way we configure them/use them - WIth that I mean both use info or int?

zroubalik: Unfortunately not in the moment, as Metrics Server internally uses https://github.com/kubernetes-sigs/custom-metrics-apiserver and we depends on it's way of logging (int based)

tomkerkhove: Ok, no problem!

PR #24, opened 2020-02-27
add option for users to insert labels on KEDA operator deployment

dimberman:
When creating networkPolicies across namespaces, it's useful to assign labels to the keda operator (the example here being that in the postgreSQL-scaler, we need to access a postgres service in a separate namespace).

### DIFF
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -14,6 +14,8 @@ watchNamespace: ""
 imagePullSecrets: []
 operatorName: keda-operator
 
+extraLabels: ""
+
 customResourceDefinition:
   create: true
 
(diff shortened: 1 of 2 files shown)


keda/values.yaml
```diff
@@ -14,6 +14,8 @@ watchNamespace: ""
 imagePullSecrets: []
 operatorName: keda-operator
 
+extraLabels: ""
```
tomkerkhove: Just a suggestion - Maybe call it `customLabels` or `additionalLabels`?

tomkerkhove:
Thanks!

jeffhollan:
Do note though that without running the command in the README to create the new tar.gz it won’t take effect. It’s fine as the changes are in the chart now and when we cut 1.3 they will be in, but if you wanted to make sure it was in 1.2 or a 1.2.1 version of chart you’d have to include the changes from running README

tomkerkhove:
Sounds good to me

Tom Kerkhove
Azure Architect at Codit, Microsoft Azure MVP

________________________________

PR #27, opened 2020-03-10
Provide capability to specify annotations in Helm chart

### LINKED ISSUE #26: Allow pod annotations via Chart
We would like to automatically add annotations to the controller and metrics pod. This currently does not seem possible via the helm chart. Please add this capability.

Our specific use case would be for kiam instance role annotations for AWS auth.

tomkerkhove:

Closes #26

### DIFF
--- a/keda/templates/12-keda-deployment.yaml
+++ b/keda/templates/12-keda-deployment.yaml
@@ -12,9 +12,9 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 {{ tpl .Values.additionalLabels . | indent 4}}
-  {{- if .Values.annotations.operator }}
+  {{- if .Values.podAnnotations.keda }}
   annotations:
-  {{ tpl .Values.annotations.operator . | indent 4}}
+  {{ tpl .Values.podAnnotations.keda . | indent 4}}
   {{- end }}
 spec:
   replicas: {{ .Values.replicaCount }}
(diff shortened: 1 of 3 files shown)


ncrothe:
That should work. Comparing this to the image values, I would suggest using annotation.keda and annotation.metricsAdapter for consistency. Also, I would suggest podAnnotation instead of plain annotation as the chart might end up with serviceAnnotations etc. at some point.


Looking at the structure of the values section, the multi-component nature of the chart and other charts I have seen, I do, however, think you might want to consider switching from a <property>.<component> structure to <component>.<property> eventually.

I.e. from
```
annotations:
  operator:
  apiserver:
```
etc. to something like
```
operator:
  image:
  replicaCount:
  podAnnotations:
metrics:
  image:
  replicaCount:
  podAnnotations:
```
Though this does go well beyond the scope of an MR like this and would also break existing deployments. But something to consider for a future version bump.

ncrothe:
Sorry, did not realise github ate the angular brackets. Fixed my suggestion to reflect. @tomkerkhove

ncrothe:
@ahmelsayed Is it possible to make a timely 1.3.1 release of the chart? Because this fix would go really well with the annotation-based podIdentity stuff in the 1.3.0 release of the operator.

ahmelsayed:
@ncrothe Done :)

ncrothe:
As pointed out in https://github.com/kedacore/keda/issues/706 the annotations ended up in the wrong place. The are currently in the deployments' metadata, but need to go into the spec's metadata, i.e. to spec.template.metadata.

PR #30, opened 2020-03-13
KEDA-release-1.3.0

SatishRanjan:


tomkerkhove:
Just a note for next releases, we only have to bump the `appVersion` and not `version` unless something really changed in the chart.

jeffhollan:
In this case though since I believe a few pull requests that introduced new params and stuff I imagine version was appropriate?

tomkerkhove:
That's right, is more of a general remark/question, sorry!

Issue #35, opened 2020-04-01
Annotations are not on spec-level for our deployments


tomkerkhove:
> The annotations for using kube2iam are in the wrong location in helm chart package...
> 
> 12-keda-deployment.yaml
> 22-metrics-deployment.yaml
> should be at the spec level in the deployment

Reported by @bmacauley-reward

Relates to https://github.com/kedacore/keda/issues/706

tomkerkhove:
Fixed via #36

PR #36, opened 2020-04-01
Move podAnnotations to deployment specs

ncrothe:
After discussions in https://github.com/kedacore/charts/issues/35 and https://github.com/kedacore/keda/issues/706, this moves the podAnnotations to the correct place in deployment.spec.template


ncrothe:
@tomkerkhove I switched from tpl to toYaml to make this work with typical helm values. Was there any particular reason you used tpl in your PR?

tomkerkhove:
No not perse, thanks for the contribution!

PR #43, opened 2020-04-20
Allow metric adapter to serve a single namespace

amirschw:
Pass `WATCH_NAMESPACE` environment variable as a value to allow metric adapter to serve a single namespace.

Signed-off-by: amirschw <24677563+amirschw@users.noreply.github.com>

### DIFF
--- a/keda/Chart.yaml
+++ b/keda/Chart.yaml
@@ -4,7 +4,7 @@ description: Event-based autoscaler for workloads on Kubernetes
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.3.2
+version: 1.3.3
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
--- a/keda/templates/22-metrics-deployment.yaml
+++ b/keda/templates/22-metrics-deployment.yaml
@@ -45,7 +45,7 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           env:
             - name: WATCH_NAMESPACE
-              value: ""
+              value: {{ .Values.watchNamespace }}
           args:
           - /usr/local/bin/keda-adapter
           - --secure-port=6443


zroubalik:
@amirschw what is your intention behind this change please?

oronboni:
@zroubalik scenario:
Each namespace is different development team.
Each team use different azure identity.
The goal is to deploy KEDA for each of the namespaces and every one of them will be with his unique identity and with his own operator and metrics server.

zroubalik:
@oronboni makes sense. Unfortunately there could be only one metrics server in kuberentes cluster. So at least metrics server have to be shared.

oronboni:
@zroubalik )-: question:
Step #1: I will deploy keda-operator in different namespaces (with different azure identity)
Step #2:  deploy one metrics server without watch_namespace.
Do all the request for Azure will use the keda-operator-metrics-server identity ?

amirschw:
> @oronboni makes sense. Unfortunately there could be only one external (KEDA) metrics server in kuberentes cluster. So at least metrics server have to be shared.

@zroubalik If I understand correctly, this means that we can’t have multiple namespace scoped operators in a cluster using this chart, since every release will install a cluster-wide metrics adapter in a different namespace. Please correct me if I got it wrong.

oronboni:
@zroubalik you are correct if I am using the current helm chart without changes. 

But I checked the following scenario:
1. Deployed in namespace A and in namespace B.
2. Removed manually keda-operator-metric-server from namespace B.

What I saw is keda-operator-metric-server in namespace A handle both of the keda-operators (the one in namespace A and the one in namespace b).

And I answer to myself (checked the identity) the Azure identity is the one the relate to  keda-operator-metric-server (meaning cannot authenticate to Azure with different Azure identities).

Question:
Do you know if you plan to support in this option in the future ?

zroubalik:
> > @oronboni makes sense. Unfortunately there could be only one external (KEDA) metrics server in kuberentes cluster. So at least metrics server have to be shared.
> 
> @zroubalik If I understand correctly, this means that we can’t have multiple namespace scoped operators in a cluster using this chart, since every release will install a cluster-wide metrics adapter in a different namespace. Please correct me if I got it wrong.

@amirschw there could by one one (KEDA) External Metrics server in the cluster, so the last one you will install will overwrite the previous one. You can have multiple KEDA operators in the cluster for handling the scaling logic though.

zroubalik:
@oronboni yeah, that's the case. It is a limitation of https://github.com/kubernetes-sigs/custom-metrics-apiserver

To answer your question, it would be better to create a separate issue for this as it would require significant changes in KEDA (if it is even possible).

amirschw:
Thanks @zroubalik, I am closing this then.

elutsky:
Hi @zroubalik 
I am working with @oronboni.
Can you please to explain a bit why it is required to re-design the keda?

Just to make sure we are aligned - what we are trying to achieve is:
1. We have different namespaces (teams), accessing external (azure resources) metrics
2. we would like those teams each to use it's own AAD Identity - to ensure least privilege access
3. we actually don't care to have a shared metrics server - we just need the external requests for the metrics to be done by per-namespace identities.

If you think it is impossible with current Keda architecture - can you explain which workaround can be taken, and to provide some draft of estimation for this effort - we'd consider the contribution here.

zroubalik:
Hi @elutsky could you please create separate issue with your requirements and observations, so it could be discussed. Closed PR is not the best place for other people to contribute. Thanks!

PR #45, opened 2020-04-21
Ship KEDA v1.4.0 chart

### LINKED ISSUE #42: Version should be updated to 1.4.0 not to 1.3.2
@SatishRanjan when you did release, you haven't updated version to 1.4.0. The PR should be checked and be approved by at least one person to avoid these kind of problems. 

https://github.com/kedacore/charts/blob/345ae4384beaef78aa4a8d9e56edbb0dbde62a7d/keda/Chart.yaml#L7-L11

tomkerkhove:
Ship KEDA v1.4.0 chart

Closes #42


docs/index.yaml
```diff
@@ -3,7 +3,39 @@ entries:
...
+    icon: https://raw.githubusercontent.com/kedacore/keda/master/images/keda-logo-500x500-white.png
+    maintainers:
+    - email: ahmels@microsoft.com
+      name: Ahmed ElSayed
+    name: keda
+    sources:
+    - https://github.com/kedacore/keda
+    urls:
+    - https://kedacore.github.io/charts/keda-1.3.3.tgz
+    version: 1.3.3
```
tomkerkhove, line 35: Actually I'm not sure if we should even ship this since it contains the Kubernetes resources but just for 1.3.0.

Either we need to go back to 1.3.1 branch and create it from there or just bite the bullet and ignore the fact that 1.3.2 was a release by accident and only ship 1.4?

zroubalik: Yeah, I'd stick to the easier solution.

tomkerkhove: Updated.

PR #52, opened 2020-06-10
Fix podAnnotations for keda deployment

yuriipolishchuk:
I've faced an issue with setting up kiam podAnnotation for keda deployment.

Got the following error:

```
Error: template: keda/templates/12-keda-deployment.yaml:30:21: executing "keda/templates/12-keda-deployment.yaml" at <.Values.podAnnotations.keda>: wrong type for value; expected string; got map[string]interface {}
```

For metricsAdapter everything worked as expected.

So replacing `tpl` function with `toYaml` must resolve the issue.


tomkerkhove:
Thanks!

PR #54, opened 2020-06-17
Add serviceAccount.annotations to enable EKS IAM roles

jasonstitt:
Currently, the chart supports pod annotations which enables the kiam/kube2iam case. This adds the option of a serviceaccount annotation which enables the EKS service account IAM roles case, so that `identityOwner: operator` can be used successfully with EKS.


tomkerkhove:
Thanks @jasonstitt !

PR #56, opened 2020-07-01
Provide support for authentication with secrets from HashiCorp Vault

nisan270390:
Added option to add HashiCorp Vault certificates to communicate over TLS and an option to add extra environment variables that will be passed onto Keda operator and metrics api service (like VAULT_CACERT, VAULT_TOKEN etc.)

kedacore/keda#890


tomkerkhove:
This feature has not been shipped yet, can you please remove the new version for now please?

We'll do that once the PR is ready to merge and determine what most appropriate version is.

PR #57, opened 2020-07-07
Provide KEDA chart v1.5.0

tomkerkhove:
Provide KEDA chart v1.5.0


docs/index.yaml
```diff
@@ -1,9 +1,29 @@
...
+    name: keda
+    sources:
+    - https://github.com/kedacore/keda
+    urls:
+    - https://kedacore.github.io/charts/keda-1.5.0.tgz
+    version: 1.5.0
   - apiVersion: v1
     appVersion: 1.4.1
-    created: "2020-06-11T08:05:23.8025036+02:00"
+    created: "2020-07-08T10:57:25.2039794+02:00"
```
zroubalik, line 26: Pardon my ingorance and lack of HELM knowledge, but is there a way how to not update timestamps for old releases? Or is this something that's ususal in HELM?

tomkerkhove: This is due to `helm repo index docs` but will check if it has an option to leave it be

tomkerkhove: https://github.com/helm/helm/issues/7363

tomkerkhove: There is no option to skip that unfortunately - https://helm.sh/docs/helm/helm_repo_index/

zroubalik: ok, thanks!

Issue #33, opened 2020-03-26
Use `apiextensions.k8s.io/v1` instead of `apiextensions.k8s.io/v1beta1`


tomkerkhove:
Use `apiextensions.k8s.io/v1` instead of `apiextensions.k8s.io/v1beta1` since it's deprecated as of Kubernetes 1.16.

tomkerkhove:
Closed in #60

PR #60, opened 2020-08-09
Updating for v2 and adding prometheus options for scraping

tbickford:
This PR updates the chart for V2. I also removed the CRD's from the templates directory and created a template file to reference the crds directory - this should allow for the use of both Helm 2 and Helm 3.  

I add the `prometheus.enabled` option that allows for configuring the metrics port and path along with an option of also creating a `PodMonitor` for the Prometheus Operator.

I am wondering if there be a v2 branch that this PR should instead go against?


keda/values.yaml
```diff
@@ -119,4 +119,16 @@ priorityClassName: ""
 ## Extra environment variables that will be passed onto Keda operator and metrics api service
 env:
 # - name: ENV_NAME
-#   value: 'ENV-VALUE'
\ No newline at end of file
+#   value: 'ENV-VALUE'
+
+prometheus:
```
tomkerkhove, line 124: Do we turn them on or off by default?

keda/values.yaml
```diff
@@ -3,8 +3,8 @@
 # Declare variables to be passed into your templates.
 
 image:
-  keda: docker.io/kedacore/keda:1.5.0
-  metricsAdapter: docker.io/kedacore/keda-metrics-adapter:1.5.0
+  keda: docker.io/kedacore/keda:2.0.0-alpha1
```
tomkerkhove, line 6: Hm, why don't we have a GitHub release for this? 🤔

keda/values.yaml
tbickford: Good catch. That was an oversight on my end - they were turned on for testing purposes.

keda/values.yaml
tbickford: Should I set this PR as a draft until v2 is released?

tomkerkhove: I'll leave it up to @zroubalik but might be good to merge since we don't reindex the chart yet or use a v2 branch

zroubalik: What about we merge this one once v2 alpha is out?

tomkerkhove: It already is https://hub.docker.com/layers/kedacore/keda/2.0.0-alpha1/images/sha256-138b54a56b0b973597e2763e1a431a17eee11081e7a9ff07ec6eaa3085e19044?context=explore (somehow)

zroubalik: I don't mean the image, but the actual release :)

tomkerkhove: Oh ok :D

tomkerkhove:
Can we rebase this since it has been changes since then please?

Might have some conflicting changes with https://github.com/kedacore/charts/pull/69 so we'll need to 👀 that.

tbickford:
The Helm 3 lint check had me change the CRD apiVersoin from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1` but it looks like that is now causing an issue when I try to install it locally: 

```
Error: failed to install CRD crds/scaledjobs.keda.sh.yaml: CustomResourceDefinition.apiextensions.k8s.io "scaledjobs.keda.sh" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
```
It looks like there are some changes that would be needed to adapt to the updated api - https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/

Can we update KEDA to use this new schema?

tbickford:
The other linting problem that appears to be introduced in Helm 3.3 is: 
```
[ERROR] templates/21-metrics-clusterrolebinding.yaml: object name does not conform to Kubernetes naming requirements: "keda-operator:system:auth-delegator": invalid metadata name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 253
```

keda/crds/scaledjobs.keda.sh.yaml
```diff
@@ -1,8 +1,9 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
tomkerkhove, line 1: I can't find the issue exactly but think we've discussed not to bump the version here as it enforced Kubernetes 1.16, right @zroubalik?

tomkerkhove: Tracked on https://github.com/kedacore/charts/issues/33 here, but there was another issue as well

tomkerkhove:
> The Helm 3 lint check had me change the CRD apiVersion (see this [kedacore/charts/runs/1147109297?check_suite_focus=true](https://github.com/kedacore/charts/runs/1147109297?check_suite_focus=true)) from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1` but it looks like that is now causing an issue when I try to install it locally:
> 
> ```
> Error: failed to install CRD crds/scaledjobs.keda.sh.yaml: CustomResourceDefinition.apiextensions.k8s.io "scaledjobs.keda.sh" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
> ```
> 
> It looks like there are some changes that would be needed to adapt to the updated api - [kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/)
> 
> Can we update KEDA to use this new schema?
> 
> These linting errors are only appearing in the most recent version of Helm 3 (`3.3.3`). I did not encounter the problems locally using version `3.2.3`

Woops, missed this comment before reviewing. We decided to ignore it for now since we dont' want to force users to use Kubernetes 1.16 - Sorry should've told you.

If using Helm 3.2.3 doesn't show this we should maybe change our CI to use that?

zroubalik:
Yeah, we discussed it here: https://github.com/kedacore/keda/issues/701#issuecomment-693331121

We can regenerate the CRDs for `apiextensions.k8s.io/v1` and we should document that KEDA is supported only on k8s 1.16+

tbickford:
@zroubalik I merged in the changes you did for the CRD API changes today.

zroubalik:
@tbickford awesome, thanks. Just let us know, when this PR is ready :)

tbickford:
I am still trying to dig into:
```
templates/21-metrics-clusterrolebinding.yaml: object name does not conform to Kubernetes naming requirements: "keda-operator:system:auth-delegator": invalid metadata name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 253
```

If we use a prior minor version of helm3 the lint command should work.

zroubalik:
> I am still trying to dig into:
> 
> ```
> templates/21-metrics-clusterrolebinding.yaml: object name does not conform to Kubernetes naming requirements: "keda-operator:system:auth-delegator": invalid metadata name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 253
> ```
> 
> If we use a prior minor version of helm3 the lint command should work.

If that's the biggest problem, I think that we could rename that cluster role binding to conform the requirements. But we should rename that in the keda repo first, so the resources are in sync

tbickford:
@zroubalik FYI - https://github.com/helm/helm/issues/8835

zroubalik:
> @zroubalik I merged in the changes you did for the CRD API changes today.

@tbickford just be sure that you are mergin in the patched version of ScaledJob CRD, ie the output of `make release` or `make deploy`.

The ScaledObject CRD is being patched due to the problems with the required fields in the generated resources. See https://github.com/kedacore/keda/pull/1208 and https://github.com/kedacore/keda/issues/927

tomkerkhove:
@tbickford I will merge this with the current error so we can include it in the RC; would you mind creating an issue for your follow-up please?

@zroubalik I'll take a look at the CRD.

PR #63, opened 2020-08-19
Use apiextensions.k8s.io/v1 instead of apiextensions.k8s.io/v1beta1

tomkerkhove:
Use apiextensions.k8s.io/v1 instead of apiextensions.k8s.io/v1beta1 to fix linting issue

### DIFF
--- a/keda/crds/keda.k8s.io_scaledobjects_crd.yaml
+++ b/keda/crds/keda.k8s.io_scaledobjects_crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: scaledobjects.keda.k8s.io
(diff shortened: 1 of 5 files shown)


zroubalik:
@tomkerkhove is this something for v2? Then we should update the CRDs in the helm, ie. migrate those from kedacore/keda

tomkerkhove:
This is mainly to fix the linting issue which is failing on this, but it's only supported as of v1.16 (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) so I'm wondering if this is even good to go or not.

zroubalik:
Yeah, this issue is decribed here: https://github.com/kedacore/keda/issues/701

I think that it is not needed to change it now, `v1beta1` is still going to work. If we do the change we would lost support on k8s < 1.16 as you mentioned.

We can change this later as part of some 2.x release, it is not as big change is it could look.

tomkerkhove:
Sounds good to me, we'll just ignore that one error for now then.

What about the name of the cluster role binding?

zroubalik:
> What about the name of the cluster role binding?

The name is taken from the Cluster role name it is referring to `system:auth-delegator`.

But it is ok to rename it, if you do so, please change it in the kedacore/keda repo as well, just for the consistency.

tomkerkhove:
Let's just keep things as is. This is nothing major.

PR #67, opened 2020-09-11
2.0.0 Beta release

tomkerkhove:


jeffhollan (approved):

jeffhollan:
Actually should we do this in a way so folks have to explicitly set version to get 2.0 and so if just running command it'll pull in 1.5?

tomkerkhove:
-beta will only show up if they include --devel when listing

PR #69, opened 2020-09-19
Update CRDs for v2 + fix operator commands/arguments

mboutet:
Changes:
* Update CRDs for v2
* Fix incorrect command and arguments for keda deployment

I tested my changes on a fresh AKS cluster v1.18.6 and the operator now deploys correctly.

Fixes https://github.com/kedacore/keda/issues/1175

Also, do we create a new release of the chart?

PR #70, opened 2020-09-21
Ship KEDA 2.0.0-beta1.1

tomkerkhove:

Shipping #69


zroubalik:
`2.0.1-beta` ? If you want to release a new beta version, the correct name should be something like `2.0.0-beta2`

zroubalik:
But if we are about to release `2.0.0-beta2` charts, I think that we should release `2.0.0-beta2` in `kedacore/keda` as well.

tomkerkhove:
Actually it's just a patch of the Helm chart 
New beta might make sense but not sure if we should go to kedacore/keda since nothing has changed? 🤔

zroubalik:
Yeah, but you are patching (thus updating) the `Beta` version, not `2.0.0`. Therefore it should be probably named `2.0.0-beta1.1`

tomkerkhove:
Makes sense, shall I use that?

zroubalik (approved):
LGTM

PR #78, opened 2020-10-08
Update images' tag (from beta to rc)

mboutet:
The tags in the `values.yaml` were still those for the beta release. This PR updates the tags to be those of the RC release.

Signed-off-by: Maxence Boutet <mboutet@explorance.com>

### DIFF
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -3,8 +3,8 @@
 # Declare variables to be passed into your templates.
 
 image:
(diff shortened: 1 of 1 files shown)


zroubalik (approved):

zroubalik:
Thanks!

tomkerkhove:
This should not be a problem as it should default to the app version but thanks for the alignment!

PR #84, opened 2020-10-19
Change cluster role binding name for metrics adapter

tomkerkhove:

Change cluster role binding name for metrics adapter to comply with Helm naming convention.


tomkerkhove:
@zroubalik If you agree on this I'll send a PR to kedacore/keda as well.

zroubalik:
We talked about this with @tbickford and it shouldn't be needed to do this change. That's a bug in Helm if I am not mistaken.

tomkerkhove:
So, what do we do with this one? Do we include it in 2.0 or abandon the PR?

zroubalik:
@tomkerkhove let's double check on standup and with @tbickford

ahmelsayed (approved):

tomkerkhove:
@zroubalik We forgot to check but let's not inclulde it for 2.0

zroubalik:
@tomkerkhove agree, we can reiterate later

PR #94, opened 2020-11-03
remove helm 2 and index 2.0.0-rc3

jeffhollan:
Signed-off-by: Jeff Hollan <jeffhollan@outlook.com>

Resolves #88

PR #95, opened 2020-11-03
Package KEDA 2.0 chart

jeffhollan:
Figured while I was at it may as well open the PR for 2.0.0 chart

Related to #94


tomkerkhove:
Using https://github.com/kedacore/charts/pull/97 instead of this one since there were merge conflicts

PR #101, opened 2020-11-16
Change Helm Chart API version to v2

allenmunC1:
Fixes issue with argocd unable to install KEDA as argocd by default uses helm2 when `apiVersion` is set to `v1` in the `Chart.yaml`

Changes
Changed `apiVersion` to `v2` in `Chart.yaml` to automatically default argocd to use helm3 when installing the chart

References
https://argoproj.github.io/argo-cd/user-guide/helm/#helm-version

### DIFF
--- a/keda/Chart.yaml
+++ b/keda/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 name: keda
 description: Event-based autoscaler for workloads on Kubernetes
 


tomkerkhove:
Thanks, this fell through the cracks unfortunately.

tomkerkhove (approved):

tomkerkhove:
@zroubalik Shall we go with 2.1 for this or hold off until the other PRs are in?

zroubalik:
@tomkerkhove not sure, I'd wait some time before 2.1. so we can even release charts 2.0.1 version if it is needed

PR #119, opened 2021-01-27
Allow read/list/get of serviceaccounts

mf-lit:
Allow the operator to read serviceaccounts.

This is required for the [aws-eks pod identity provider](https://keda.sh/docs/2.0/concepts/authentication/#eks-pod-identity-webhook-for-aws) to work. IAM roles are bound to serviceAccounts (using an annotation), and so the operator needs to be able to read the relevant serviceAccount in order to discover which role it should assume.

### DIFF
--- a/keda/templates/10-keda-clusterrole.yaml
+++ b/keda/templates/10-keda-clusterrole.yaml
@@ -26,6 +26,7 @@ rules:
   - pods
   - secrets
   - services
+  - serviceaccounts
   verbs:
   - get
   - list


ahmelsayed (approved):
LGTM

tomkerkhove (approved):

keda/templates/10-keda-clusterrole.yaml
```diff
@@ -26,6 +26,7 @@ rules:
   - pods
   - secrets
   - services
+  - serviceaccounts
```
tomkerkhove, line 29: @zroubalik I presume we'll need to backport this to https://github.com/kedacore/keda/blob/main/config/rbac/role.yaml#L23?

If so, I'll send a PR

PR #125, opened 2021-02-22
Allow define number of replicas of keda

mtparet:
Allow to define number of replicas for Keda for HA.

### Checklist

### DIFF
--- a/keda/Chart.yaml
+++ b/keda/Chart.yaml
@@ -4,7 +4,7 @@ description: Event-based autoscaler for workloads on Kubernetes
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.0
+version: 2.5.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
--- a/keda/README.md
+++ b/keda/README.md
@@ -72,6 +72,7 @@ their default values.
 | `podLabels.keda`                                           | Pod labels for KEDA operator | `{}` |
 | `podLabels.metricsAdapter`                                 | Pod labels for KEDA Metrics Adapter | `{}` |
 | `podDisruptionBudget`                                      | Capability to configure [Pod Disruption Budget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)       | `{}` |
+| `keda.replicaCount`                                      | Capability to configure number of replicas for Keda   | `1` |
 | `rbac.create`                                              | Specifies whether RBAC should be used | `true`                                        |
 | `serviceAccount.create`                                    | Specifies whether a service account should be created       | `true`                                        |
 | `serviceAccount.name`                                      | The name of the service account to use. If not set and create is true, a name is generated.      | `keda-operator` |
--- a/keda/templates/12-keda-deployment.yaml
+++ b/keda/templates/12-keda-deployment.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/name: {{ .Values.operator.name }}
     {{- include "keda.labels" . | indent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.keda.replicaCount}}
   selector:
     matchLabels:
       app: {{ .Values.operator.name }}
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -44,6 +44,9 @@ podDisruptionBudget: {}
 #  minAvailable: 1
 #  maxUnavailable: 1
 
+keda:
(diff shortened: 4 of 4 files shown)


tomkerkhove:
Let's verify if it's supported first - https://github.com/kedacore/keda-docs/pull/381#issuecomment-783340717

atiasadir:
Is there any blocker to add it ?

tomkerkhove:
Yes as it will not do anything, see https://github.com/kedacore/keda-docs/pull/381#issuecomment-783340717

atiasadir:
> Yes as it will not do anything, see [kedacore/keda-docs#381 (comment)](https://github.com/kedacore/keda-docs/pull/381#issuecomment-783340717)

In terms of performance, It will reduce a downtime during a failover.
And sounds like this is agreed on the link you sent.

So I think it will be good to have this option.

tomkerkhove:
Thought on the failover approach @kedacore/keda-maintainers ? Makes sense, but should be documented clearly then.

JorTurFer:
> Thought on the failover approach @kedacore/keda-maintainers ? Makes sense, but should be documented clearly then.

From my pov if the operator is ready to work in HA (as it is AFAIR), there is no problem in supporting the modification of the amount of instances. Of course, we should explain the that more than one instance is only for HA and failover stuffs and not for load balancing

arschles (approved):
I agree with @JorTurFer here, including that we should document the failover/HA approach. Is there any harm in merging this before we add the docs for it?

mtparet:
The doc should be good also now https://github.com/kedacore/keda-docs/pull/381. 
I have rebased the PR on last master version.

keda/README.md
```diff
@@ -72,6 +72,7 @@ their default values.
 | `podLabels.keda`                                           | Pod labels for KEDA operator | `{}` |
 | `podLabels.metricsAdapter`                                 | Pod labels for KEDA Metrics Adapter | `{}` |
 | `podDisruptionBudget`                                      | Capability to configure [Pod Disruption Budget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)       | `{}` |
+| `keda.replicaCount`                                      | Capability to configure number of replicas for Keda   | `1` |
```
JorTurFer, line 75: Please, use KEDA instead of Keda

JorTurFer (changes requested):
LGTM
Only small nits

keda/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Event-based autoscaler for workloads on Kubernetes
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.0
+version: 2.5.1
```
tomkerkhove, line 7: Let's keep the version as-is and take that through 2.6 since this is not exactly a fix
```suggestion
version: 2.5.0
```

keda/README.md
tomkerkhove, line 75: Let's align with our current structure and add a better description
```suggestion
| `operator.replicaCount`                                      | Capability to configure the number of replicas for KEDA operator.<br /><br />While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic.<br /><br />You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.| `1` |
```

keda/templates/12-keda-deployment.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/name: {{ .Values.operator.name }}
     {{- include "keda.labels" . | indent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.keda.replicaCount}}
```
tomkerkhove, line 12:
```suggestion
  replicas: {{ .Values.operator.replicaCount}}
```

keda/values.yaml
```diff
@@ -44,6 +44,9 @@ podDisruptionBudget: {}
 #  minAvailable: 1
 #  maxUnavailable: 1
 
+keda:
```
tomkerkhove, line 47: Let's move under existing `operator`

tomkerkhove (changes requested):
We should align with the existing `operator`

tomkerkhove:
I did the fixes but apparently I can't commit to the PR :( I'll need you to do them, sorry @mtparet . The majority are OK to just use the suggestion accept for https://github.com/kedacore/charts/pull/125#discussion_r763694211

arschles (approved):

tomkerkhove:
Merging to temp branch

PR #126, opened 2021-02-26
Chart: 2.1.2 | App: Keda Fixed cluster role

mlushpenko:
Signed-off-by: mlushpenko <mlushpenko@eclecticiq.com>

Based on this:
```
E0226 16:32:19.511971       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1alpha1.ClusterTriggerAuthentication: failed to list *v1alpha1.ClusterTriggerAuthentication: clustertriggerauthentications.keda.sh is forbidden: User "system:serviceaccount:kube-system:keda-operator" cannot list resource "clustertriggerauthentications" in API group "keda.sh" at the cluster scope
```


keda/templates/10-keda-clusterrole.yaml
```diff
@@ -81,6 +81,8 @@ rules:
 - apiGroups:
   - keda.sh
   resources:
```
tomkerkhove, line 83: @zroubalik I think this is fine, thoughts?

If so, would you mind sending a PR to https://github.com/kedacore/keda/blob/main/config/rbac/role.yaml#L76 @mlushpenko ?

zroubalik: Since we are renaming a resource, there might be problem when doing update of KEDA 2.1 -> 2.2 (or the version that will include this change). Because I am not sure that the "old" resource will be properly deleted? How does Helm handle this?

tomkerkhove: It does upserts AFAIK so that should be OK. Should we port this back to the core?

zroubalik: OK! We should add a note to the Release Notes about this change. (Users who install through yaml manifest need to delete the old ClusterRole manually.

zroubalik: OMG, sorry please disregard my previous comments. I thought this PR is about renaming the ClusterRole.

This change is absolutely OKay, and should be backported to core

zroubalik (approved):

keda/templates/10-keda-clusterrole.yaml
mlushpenko: > https://github.com/kedacore/keda/blob/main/config/rbac/role.yaml#L76

@tomkerkhove  is this good https://github.com/kedacore/keda/pull/1645?

tomkerkhove: Certainly, thanks!

tomkerkhove (approved):

PR #114, opened 2020-12-27
Enabled extravolumes for both Keda and Metrics server deployment

HansK-p:
Enabled addition of extra volume mounts for both the Keda and the Metrics server deployment. This is needed if there is a need to run Metrics server with a read only root file system, hopefully there are other use cases where this can be useful too.

Ref #113

PR #146, opened 2021-04-26
helm: add extra volumes and volume mounts configuration

### LINKED ISSUE #113: Solve issue with deploying Keda using "securityContext.readOnlyRootFilesystem: true"
keda-operator-metrics-apiserver refuses to start when I install Keda using Helm 3 with the following values.yaml file:
<pre>
securityContext:
  capabilities:
    drop:
    - ALL
  readOnlyRootFilesystem: true # The setting that creates issues
  runAsNonRoot: true
  runAsUser: 1000
</pre>

Earlier we have "solved" the problem by:
* using "helm template" to create a helm template
* use kustomize to patch the keda-operator-metrics-apiserver deployment so that two emptydir volumes are mounted as /tmp and /apiserver.local.config
* apply the patched yaml files

But we plan on making some changes to our deployment process so that this extra step makes the deployment process more cumbersome.

It would be good to be able to solve this issue using some kind of Helm variable (possibly using Helm variables with a name similar to additionalVolumes & additionalVolumeMounts).

It's not extremely important that these pods have a read only file system, but we plan on adding policies enforcing the use of read-only pod root file systems, and it would be good not to have to create an exception for the Keda deployment/avoid the extra complexity of patching these Helm generated yaml files.

t3mi:
Continuation of the work [PR#114](https://github.com/kedacore/charts/pull/114)
credits to @HansK-p
### Checklist

Fixes #113


tomkerkhove (approved):
LGTM, thank you!

PR #151, opened 2021-05-17
Default image tag and remove 'v' prefix

### LINKED ISSUE #150: Allow the http-add-on chart to use the 'latest' tag
Currently the HTTP add on chart defaults to a [`vX.Y.Z` tag](https://github.com/kedacore/charts/blob/914c60da8d2805be03f562b3720307683af4d4cd/http-add-on/templates/deployment.yaml#L53). We should allow users to install the `latest` tag as well. This

### Use-Case

Users who would like to stay on the latest stable release may want to just use `latest` (along with an `ImagePullPolicy` of `Always`) to stay up to date.

### Specification

- [ ] The [deployment's image name](https://github.com/kedacore/charts/blob/914c60da8d2805be03f562b3720307683af4d4cd/http-add-on/templates/deployment.yaml#L53) is fixed to not always prefix with `v`
- [ ] The [values.yaml](https://github.com/kedacore/charts/blob/master/http-add-on/values.yaml#L20-L22) file is changed to add a `tag` field, similar to how KEDA's [values.yaml](https://github.com/kedacore/charts/blob/master/keda/values.yaml#L9) does it

cc/ @khaosdoctor

khaosdoctor:
Removes the 'v' prefix from images and defaults the image tag to latest or the chart version

### Checklist

Fixes #150
Related to https://github.com/kedacore/http-add-on/pull/167 and https://github.com/kedacore/http-add-on/pull/125

### DIFF
--- a/http-add-on/README.md
+++ b/http-add-on/README.md
@@ -82,7 +82,7 @@ their default values.
 
 | Parameter                                                  | Description                               | Default                                         |
 |:-----------------------------------------------------------|:------------------------------------------|:------------------------------------------------|
-| `images.tag`                                               | Image tag for the http add on             | `latest`                              |
+| `images.tag`                                               | Image tag for the http add on             | None, it uses Helm chart's app version as a default                              |
 | `images.operator`                                          | Image name for the operator image component | `ghcr.io/kedacore/http-add-on-operator:latest` |
 | `images.interceptor`                                       | Image name for the interceptor image component | `ghcr.io/kedacore/http-add-on-interceptor:latest` |
 | `images.scaler`                                            | Image name for the scaler image component | `ghcr.io/kedacore/http-add-on-scaler:latest` |
--- a/http-add-on/README.md
+++ b/http-add-on/README.md
@@ -82,11 +82,10 @@ their default values.
 
 | Parameter                                                  | Description                               | Default                                         |
 |:-----------------------------------------------------------|:------------------------------------------|:------------------------------------------------|
-| `version`                                                  | Image tag for the http add on             | The latest release                              |
+| `images.tag`                                               | Image tag for the http add on             | `latest`                              |
 | `images.operator`                                          | Image name for the operator image component | `ghcr.io/kedacore/http-add-on-operator:latest` |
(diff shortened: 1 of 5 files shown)


tomkerkhove (changes requested):
LGTM, can you update the README with your new config value please?

arschles (approved):
@khaosdoctor thanks!

khaosdoctor:
@tomkerkhove done!

arschles (approved):

http-add-on/values.yaml
```diff
@@ -17,6 +15,7 @@ operator:
   port: 8443
 
 images:
+  tag: latest
```
tomkerkhove, line 18: Since we'd prefer it to default to the app version, we should just leave this empty here
```suggestion
  tag:
```

http-add-on/README.md
```diff
@@ -82,12 +82,11 @@ their default values.
 
 | Parameter                                                  | Description                               | Default                                         |
 |:-----------------------------------------------------------|:------------------------------------------|:------------------------------------------------|
-| `version`                                                  | Image tag for the http add on             | The latest release                              |
+| `images.tag`                                               | Image tag for the http add on             | `latest`                              |
```
tomkerkhove, line 85: Since we'd prefer it to default to the app version, we should just leave this empty here
```suggestion
| `images.tag`                                               | Image tag for the http add on             | None, it used Helm chart's app version as a default                              |
```

http-add-on/templates/deployment.yaml
```diff
@@ -3,12 +3,12 @@ kind: Deployment
 metadata:
   labels:
     control-plane: controller-manager
-    httpscaledobjects.http.keda.sh/version: {{ .Values.version }}
+    httpscaledobjects.http.keda.sh/version: {{ .Chart.AppVersion }}
```
tomkerkhove, line 6: What if somebody overrides it with ´images.tag´?

(applies to all of these though)

tomkerkhove (changes requested):
LGTM, but:
- We just need to leave out the default image tag value so that it defaults to the app's chart version.
- Labels should check both values as well

khaosdoctor:
Done!

tomkerkhove (approved):

PR #174, opened 2021-08-03
fix(metrics-service): expose metrics port when metrics are enabled

### LINKED ISSUE #173: Metrics are not coming into Prometheus when metrics are enabled
Metrics are not coming into Prometheus when metrics are enabled in `prometheus.metricServer.enabled`

## Expected Behavior
exposed metrics should propagate

## Actual Behavior
expose metrics are not propagating because metrics port `9022` is not exposed by the service

  1. Enable metrics `prometheus.metricServer.enabled = true` in `keda/values.yaml`
  2. Modify scrape configs in prometheus (VictoriaMetrics is used in my configuration but could be applied to `<scrape_config>` in prometheus
  3. Query one of the sample metrics `keda_metrics_adapter_scaled_object_errors`

## Specifications

  - **KEDA Version:** v2.3.0 chart: v2.3.2
  - **Platform & Version:** *Please elaborate*
  - **Kubernetes Version:** v1.18 (EKS)
  - **Scaler(s):** *Please elaborate*

samarara:
Signed-off-by: Sandes de Silva <sandes.de.silva@roadmunk.com>

When `prometheus.metricServer.enabled` is true, this port must be exposed by the service in order for Prometheus or VictoriaMetrics scrape configurations to add this as a target. 

### Checklist

Fixes #173


ahmelsayed (approved):
LGTM. @tomkerkhove not sure if changes in `docs/` directory should be part of this PR or the future release PR?

tomkerkhove:
@ahmelsayed It's OK to keep as-is since we'll ship a new version tomorrow!

keda/README.md
```diff
@@ -107,6 +108,7 @@ their default values.
 | `prometheus.metricServer.podMonitor.additionalLabels`      | Additional labels to add for metric server using podMonitor crd (prometheus operator) | `{}`
 | `prometheus.operator.enabled`                              | Enable keda operator prometheus metrics expose | `false`
 | `prometheus.operator.port`                                 | HTTP port used for exposing keda operator prometheus metrics | `9022`
+| `prometheus.operator.port`                                 | HTTP port name used for exposing keda operator prometheus metrics | `metrics`
```
tomkerkhove, line 111:
```suggestion
| `prometheus.operator.portName`                                 | HTTP port name used for exposing keda operator prometheus metrics | `metrics`
```

keda/values.yaml
```diff
@@ -171,6 +172,7 @@ prometheus:
   operator:
     enabled: false
     port: 8080
+    portName: metrics
```
tomkerkhove, line 175: As far as I see, this is not being used - Is this on purpose?

tomkerkhove (commented):
Added a few remarks - Can you also please revert the new chart version please?

tomkerkhove:
@zroubalik Do you want this in core repo as well?

keda/values.yaml
samarara: Yeah. I left that out because I'm not sure if the operator metrics could also be exposed. From the docs it seemed like metrics-apiserver metrics is the only one needed: https://keda.sh/docs/2.1/operate/prometheus/

I can certainly add this to the `keda-deployment` config but then we would need to add a service to the keda operator in order to expose these. (afaik)

samarara: Or I can remove `portName` and the related docs change for the time being @tomkerkhove. What would you suggest?

tomkerkhove: I like the consistency, but if it's doesn't do anything I'd suggest to leave it out until we need it - What do you think?

zroubalik: Exposing metrics from Operator would be nice to have as well. It doesn't have to be solved in this PR of course :)

zroubalik: The whole prom metrics stack in KEDA (both Operator & Metrics Server) deserves some unification and refactoring.

samarara:
@tomkerkhove @zroubalik I removed references to `prometheus.portName` (mistakenly named `port`) from the `values.yaml` and `README.md`. I'll add this to an operator service in a following PR.

tomkerkhove:
Thank you @samarara - Would you mind reverting the new version as well please?

samarara:
As `2.3.2` @tomkerkhove ?

tomkerkhove:
> ~As `2.3.2` @tomkerkhove ?~
> Oh I see what you mean. I forgot to package it.

The package, chart version and the updated docs index indeed - Thanks for fixing!

PR #177, opened 2021-08-06
feat: expose operator metrics via a service

samarara:
Signed-off-by: Sandes de Silva <sandes.de.silva@roadmunk.com>

- add a service for operator that forwards the target port to `metrics-addr` arg
- expose operator metrics at `http://keda-operator.<namespace>.svc.cluster.local/metrics`

### Checklist

Fixes #

### DIFF
--- a/keda/README.md
+++ b/keda/README.md
@@ -92,11 +92,14 @@ their default values.
 | `priorityClassName`                                        | Pod priority for KEDA Operator and Metrics Adapter ([docs](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/)) | `` |
 | `env`                                                      | Additional environment variables that will be passed onto KEDA operator and metrics api service | `` |
 | `http.timeout` | The default HTTP timeout to use for all scalers that use raw HTTP clients (some scalers use SDKs to access target services. These have built-in HTTP clients, and the timeout does not necessarily apply to them) | `` |
-| `service.annotations`                                      | Annotations to add the KEDA Metric Server service | `{}`                                        |
-| `service.portHttp`                                         | Service HTTP port for KEDA Metric Server service | `80`                                        |
-| `service.portHttpTarget`                                   | Service HTTP port for KEDA Metric Server container | `8080`                                        |
-| `service.portHttps`                                        | HTTPS port for KEDA Metric Server service | `443`                                        |
-| `service.portHttpsTarget`                                  | HTTPS port for KEDA Metric Server container | `6443`                                        |
+| `service.operator.annotations`                             | Annotations to add the KEDA Operator service | `{}`                                        |
+| `service.operator.portHttp`                                | Service HTTP port for KEDA Operator service | `80`                                        |
+| `service.operator.portHttpTarget`                          | Service HTTP port for KEDA Operator container | `8080`                                        |
+| `service.metricServer.annotations`                         | Annotations to add the KEDA Metric Server service | `{}`                                        |
+| `service.metricServer.portHttp`                            | Service HTTP port for KEDA Metric Server service | `80`                                        |
+| `service.metricServer.portHttpTarget`                      | Service HTTP port for KEDA Metric Server container | `8080`                                        |
+| `service.metricServer.portHttps`                           | HTTPS port for KEDA Metric Server service | `443`                                        |
+| `service.metricServer.portHttpsTarget`                     | HTTPS port for KEDA Metric Server container | `6443`             –                           |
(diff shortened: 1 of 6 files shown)


samarara:
@tomkerkhove I'd love your feedback on this. Since metrics for the operator is exposed at the default port it's running in, I removed `prometheus.operator.port` I also made the operator deployment a bit dynamic using port values from `service.operator`. I'll modify the `README` once I have some feedback. Thanks!

ahmelsayed (approved):

tomkerkhove (commented):
This is definately a welcomed change but I'm not sure about the breaking nature of it though.

What do you think @zroubalik? Do you want to have this in the core as well?

zroubalik (approved):

zroubalik:
> This is definately a welcomed change but I'm not sure about the breaking nature of it though.
> 
> What do you think @zroubalik? Do you want to have this in the core as well?

Good idea :)

samarara:
Great! I'll update the README and send a complete MR shortly.

samarara:
@zroubalik @tomkerkhove. Sorry for the delay. Just updated README and repackaged it. This one should be good to go. Let me know if you want any additional changes.

keda/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Event-based autoscaler for workloads on Kubernetes
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.0
+version: 2.5.0
```
tomkerkhove, line 7: Can you revert the new version please?

samarara: Ah damn!

tomkerkhove: It's best to just revert this change along with the new version & feed update.

We typically do this separately as part of a release (which is coming soon)

keda/README.md
```diff
@@ -107,7 +110,6 @@ their default values.
 | `prometheus.metricServer.podMonitor.namespace`             | Scraping namespace for metric server using podMonitor crd (prometheus operator) | ``
 | `prometheus.metricServer.podMonitor.additionalLabels`      | Additional labels to add for metric server using podMonitor crd (prometheus operator) | `{}`
 | `prometheus.operator.enabled`                              | Enable keda operator prometheus metrics expose | `false`
-| `prometheus.operator.port`                                 | HTTP port used for exposing keda operator prometheus metrics | `9022`
```
tomkerkhove, line 110: Is this on purpose?

tomkerkhove (changes requested):
Can you bring `docs/keda-2.4.0.tgz` back please?

keda/README.md
samarara: Yeah. Operator metrics aren't exposed at port 9022. They're exposed at wherever `.Values.service.operator.portHttpTarget/metrics`

samarara:
@tomkerkhove reverted `docs/keda-2.4.0.tgz`

tomkerkhove:
Thanks! I'll take a look asap so we can ship v3.0

tomkerkhove (commented):
I was wondering - Wouldn't it be possible to add the new variables next to the current ones so that it's not a breaking change?

samarara:
Which variables @tomkerkhove ?

tomkerkhove:
We are splitting service.annotations into service.operator.annotations & service.metricServer.annotations which is a breaking change.

Instead of doing that, why don't we use service.annotations when specified, otherwise we use service.operator.annotations/service.metricServer.annotations?

Then we can easily merge this in a non-breaking fashion.

(and this would apply to all, just using annotations as an example)

samarara:
We could. But that deviates from the already existing pattern we have for `logging` and `prometheus`. Also service before only applied to metrics-server. Are you suggesting we have the same set of values that can apply to both operator and metric-server services?

tomkerkhove:
No, just suggesting to be backwards compatible by keeping the old values and apply those when specified.

Otherwise it can potentially break a lot of end-users

samarara:
I can refactor it. What would happen if you specify both? Which on should take precedence?

tomkerkhove:
That's a good question but my gut says use the old variable which is less breaking, but I'm open to suggestions - I don't have a preference.

Thank you for the extra work and sorry about that!

*(feel free to mark the old fields as deprecated in the README)*

tomkerkhove:
Any chance you are able to finalize this PR @samarara?

tomkerkhove:
Closing due to inactivity

PR #197, opened 2021-11-15
Updating HTTP Addon interceptor's default total wait timeout

arschles:
This PR is a follow-on to https://github.com/kedacore/http-add-on/issues/316. It updates the default total amount of time the interceptor will wait for a backend to be available, before it fails a request. All other defaults stay the same.

### Checklist

Fixes #

cc/ @tpiperatgod


arschles:
Since kedacore/http-add-on#316 is the last issue required for the HTTP Addon's `v0.2.0` release, when this PR is merged, I will "hit the button" on the release.

tomkerkhove (approved):
As per the other comment, we typically publish a new version before shipping a new Chart version.

Given this is already published, I will merge it, but we should avoid it in the future.

arschles:
Thanks @tomkerkhove. Major human error here, my fault. I got some bug reports during the time between chart publication and release artifact publication, so I paid for my mistake 😆 

I updated the HTTP Addon release document in https://github.com/kedacore/http-add-on/pull/327 to reflect the right way to do things, so this won't happen again. There's also https://github.com/kedacore/charts/pull/202 that will ensure that releases can be cut in the HTTP Addon repository without breaking the chart.

PR #201, opened 2021-11-16
feat: Add optional `namespace` value

### LINKED ISSUE #196: Support the namespace as a value parameter
I am using keda as a dependency of my main chart and helm does not support specifying namespaces for dependencies. As a result keda is installing into the wrong namespace.

### Use-Case
```yaml
# ./example/Chart.yaml
dependencies:
- name: keda
  version: "2.4.0"
  repository: https://kedacore.github.io/charts
```

```sh
helm install example ./example --namespace example 
```

### Specification

- [ ] Keda still installs into the "keda" namespace
- [ ] I can optionall specify a different namespace for keda to install to

### Additional Info
To provide a value, such as a namespace, to a dependency I believe that your values.yaml file should define a field called `namespace: keda`.

Then throughout the resources in your chart you would use:
```yaml
metadata:
  namespace: {{ .Values.namespace }}
```

Then it would always go to the keda namespace and if I wanted it to go to a different namespace then I would run:

```sh
helm install example ./example --set keda.namesapce=whatever
```

justinmchase:
Adds an _optional_ `namespace` variable which can be specified so that keda can be installed as a helm dependency.

## Scenario 1: Set namespace via `.Release.namespace`
```sh
helm install keda1 keda --create-namespace --namespace keda1 --dry-run > withns.yaml
```
Screen Shot 2021-11-16 at 9 32 51 AM

[withns.yaml.zip](https://github.com/kedacore/charts/files/7547911/withns.yaml.zip)

## Scenario 2: Set namespace via `.Values.namespace`
```sh
helm install keda1 keda --set namespace=keda2 --dry-run > setns.yaml
```
Screen Shot 2021-11-16 at 9 38 21 AM

[setns.yaml.zip](https://github.com/kedacore/charts/files/7547916/setns.yaml.zip)

## Conclusion

This will now enable a parent chart to add keda as a dependency:

```yaml
# Chart.yaml
- name: keda
  version: "2.4.0"
  repository: https://kedacore.github.io/charts
```
```yaml
# values.yaml
keda:
  namespace: keda
```

## Optional

If we set the value to be `namespace: keda` by default it would always go into the keda namespace but it would have the downside of ignoring the `--namespace` parameter which would be a breaking change.

### Checklist

- [ ] [Read more about how you can contribute in our contribution guide](https://github.com/kedacore/charts/blob/master/CONTRIBUTING.md)
- [ ] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/master/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [ ] README is updated with new configuration values *(if applicable)*

Fixes #196

### DIFF
--- a/http-add-on/values.yaml
+++ b/http-add-on/values.yaml
@@ -1,6 +1,9 @@
 # Optionally declare a namespace via values
 # When using keda as a dependency set the value `keda.namespace` to control
-# which namespace keda will be installed in.
+# which namespace the HTTP Addon will be installed in.
+# If this is not set, the chart will use the namespace of the release.
+# If this is set, a new namespace will be created with this name
+# and the chart will use that.
 namespace:
 
 # Custom labels to add into metadata
--- a/http-add-on/values.yaml
+++ b/http-add-on/values.yaml
@@ -1,3 +1,8 @@
+# Optionally declare a namespace via values
+# When using keda as a dependency set the value `keda.namespace` to control
+# which namespace keda will be installed in.
+namespace:
+
 # Custom labels to add into metadata
 additionalLabels: ""
 
--- a/keda/templates/24-metrics-apiservice.yaml
+++ b/keda/templates/24-metrics-apiservice.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   service:
     name: {{ .Values.operator.name }}-metrics-apiserver
-    namespace: {{ .Release.Namespace }}
+  # namespace: {{ default .Release.Namespace .Values.namespace }}
   group: external.metrics.k8s.io
   version: v1beta1
   insecureSkipTLSVerify: true
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -2,6 +2,11 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# Optionally declare a namespace via values
+# When using keda as a dependency set the value `keda.namespace` to control
+# which namespace keda will be installed in.
+namespace:
+
 image:
   keda:
     repository: ghcr.io/kedacore/keda
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -5,6 +5,9 @@
 # Optionally declare a namespace via values
 # When using keda as a dependency set the value `keda.namespace` to control
 # which namespace keda will be installed in.
+# If this is not set, the chart will use the namespace of the release.
+# If this is set, a new namespace will be created with this name
+# and the chart will use that.
 namespace:
 
 image:
--- a/http-add-on/templates/deployment-interceptor.yaml
+++ b/http-add-on/templates/deployment-interceptor.yaml
@@ -15,7 +15,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
 {{ tpl .Values.additionalLabels . | indent 4}}
   name: {{ .Chart.Name }}-interceptor
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ default .Release.Namespace .Values.namespace }}
 spec:
   replicas: 1
   selector:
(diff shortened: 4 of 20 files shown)


http-add-on/values.yaml
```diff
@@ -1,3 +1,8 @@
+# Optionally declare a namespace via values
+# When using keda as a dependency set the value `keda.namespace` to control
+# which namespace keda will be installed in.
```
arschles, line 3:
```suggestion
# which namespace the HTTP Addon will be installed in.
# If this is not set, the chart will use the namespace of the release.
# If this is set, a new namespace will be created with this name
# and the chart will use that.
```

keda/templates/24-metrics-apiservice.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   service:
     name: {{ .Values.operator.name }}-metrics-apiserver
-    namespace: {{ .Release.Namespace }}
+  # namespace: {{ default .Release.Namespace .Values.namespace }}
```
arschles, line 11: it seems like we need this, shouldn't it be uncommented?

keda/values.yaml
```diff
@@ -2,6 +2,11 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# Optionally declare a namespace via values
+# When using keda as a dependency set the value `keda.namespace` to control
+# which namespace keda will be installed in.
```
arschles, line 7:
```suggestion
# which namespace keda will be installed in.
# If this is not set, the chart will use the namespace of the release.
# If this is set, a new namespace will be created with this name
# and the chart will use that.
```

arschles (changes requested):

keda/templates/24-metrics-apiservice.yaml
justinmchase: Yeah sorry, I was hoping to solicit feedback about the general approach first but there are a number of issues to  work out and one of them is that this particular line is giving me an error when I try to install it. 

It seems like its a version mismatch with Kubernetes perhaps, but it isn't working on my local machine so I commented it out for now but yes it would need to be resolved for this to go forward.

justinmchase: The other issues are the checklist items, including signed commits. I'll have to read about that a bit to make it happen.

tomkerkhove:
Feel free to let me know if you need help @justinmchase

justinmchase:
@tomkerkhove Hey, thanks, I would need some help probably.

The issue brought up here https://github.com/kedacore/charts/pull/201#discussion_r752670765 is one I'm not sure how to resolve.

I'm assuming this is a custom resource type defined by kedacore, and those would be cluster level resource right? I'm not sure why I'm getting an error.

keda/templates/24-metrics-apiservice.yaml
tomkerkhove: What was the error that you've had on this?

In terms of the commits, this should be able to help you - https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work

arschles (approved):
looks great @justinmchase !

minac:
This is great PR. We are setting up keda in our environments (great tool btw!) but need it in a different namespace and this PR does the trick. Given there's already one approval, do you think this could ship soon?

tomkerkhove:
As you can see, the PR is still in draft and the CI is failing so it depends on @justinmchase if he has the bandwidth to close it.

justinmchase:
This is still on my radar, we're planning on actually integrating Keda soon. This was part of a prototype which I proved out and we will actually adopt this soon. Sorry for the delay.

tomkerkhove:
No worries!

justinmchase:
Sorry not going to be able to finish this. Also, after working with helm dependencies quite a bit I'm fairly convinced that it would be better to just come up with a different way to install each helm chart individually rather than relying on helms dependency system. Therefore this issue is not relevant.

PR #202, opened 2021-11-18
Using immutable version of the HTTP Addon

arschles:
As of https://github.com/kedacore/http-add-on/pull/327, the HTTP Addon's release process publishes new images _before_ updating the chart. Prior to this PR, the HTTP Addon chart was using the `latest` tag for images. That means that during a release, an old chart was installing new images. This PR ensures that the chart is using immutable images, so that even when a release creates new images, the current version of the chart is using the proper ones.

### Checklist

Fixes #


tomkerkhove (approved):
Thanks a ton, a lot safer indeed!

arschles:
Thanks for the review @tomkerkhove !

PR #204, opened 2021-11-24
Update GitHub repository URL for Cosmos DB external scaler

JatinSanghvi:
Updated paths to repository and Docker image.

### Checklist


tomkerkhove (changes requested):
LGTM, but we should delete the old doc given our v0.1.0 image is still the old one, can you ever this please?

JatinSanghvi:
@tomkerkhove, can you elaborate on which file you are referring to. For clarification, I have updated the v0.1.0 release image to new path.

JorTurFer:
I prefer it like `{external|add-on}-scaler-{xxxxx}`. For the moment we have these cases but, imagine a case with IDK, 15 different elements, they will be ordered because the common part will be at the start.

Apart for that "ordering" thing, I don't have a strong opinion. What could the impact of changing http-add-on chart be? I mean, if rename it now is a drama, we could keep it as it is and adapt the others

tomkerkhove:
It would be a new chart next to the existing one, I'm afraid. But that should not be the showstopper

zroubalik:
The HTTP add on has more images, it is not just scaler:
https://github.com/orgs/kedacore/packages?repo_name=http-add-on

So I'd say it is more like:
`add-on-http-scaler`
`add-on-http-interceptor`
`add-on-foo-bar`
...

And for the external scalers:
`external-scaler-azure-cosmos-db`
`external-scaler-baz`

tomkerkhove:
Agree from my side, what do you think @JorTurFer ?

JorTurFer:
me too

zroubalik:
So let's wait what @arschles thinks about this :)

tomkerkhove (approved):
LGTM - Just waiting for @arschles his opinion

arschles:
Sorry folks, I've been out on holiday and I'm just catching up now!

I agree with you @zroubalik in your comment at https://github.com/kedacore/charts/pull/204#issuecomment-978984313. For me, it doesn't seem to make sense to have the scaler be `external-scaler-http-addon` and have the other two be `add-on-http-{operator, interceptor}`.

tomkerkhove:
Sounds good - Then the only change we have to do is rename the Helm chart for HTTP add-on unless we just keep this as-is?

PR #219, opened 2021-12-20
Allow define number of replicas of KEDA

tomkerkhove:
Allow define number of replicas of KEDA to more easily fail-over, but not use load balancing as this is not supported with operators.

### Checklist

Follow-up for #125
Related to https://github.com/kedacore/keda-docs/pull/381#issuecomment-783340717


JorTurFer (approved):

docs/artifacthub-repo.yml
```diff
@@ -0,0 +1,6 @@
+# Configuration for our Artifact Hub repository
+# See https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-repo.yml
```
JorTurFer, line 2: Care with this line because it still hits master

tomkerkhove: That's artifact hub, that's OK.

Issue #182, opened 2021-08-31
Improve the consistency in KEDA chart


JorTurFer:
Working with KEDA chart, I found that it hasn't got consistency in the names when it refers to each application, for example, sometimes the operator is named [`keda` ](https://github.com/kedacore/charts/blob/master/keda/values.yaml#L6) and other times is named [`operator`](https://github.com/kedacore/charts/blob/master/keda/values.yaml#L22). With metrics server happens the same, sometimes is called as [`metricsApiServer`](https://github.com/kedacore/charts/blob/master/keda/values.yaml#L10), or as [`metricsServer`](https://github.com/kedacore/charts/blob/master/keda/values.yaml#L25), or as [`metricsAdapter`](https://github.com/kedacore/charts/blob/master/keda/values.yaml#L33).

This is a breaking change, but maybe it could be in the next major release, and it reduces the complexity of the chart. IDK, maybe the chart values could be something like:
```yaml
operator:
  resources: {}
  tag: {}
  logging: {}
  # .....
metricsServer:
  resources: {}
  tag: {}
  logging: {}
```

IMHO, at least the naming should be unified to reduce the general complexity 
Wdyt?

tomkerkhove:
Good feedback, adding it to milestone for when we plan a new major version!

tomkerkhove:
Until then, we can change it so that it's done in a non-breaking manner but it can be quite messy so not sure if it's worth the time

JorTurFer:
Probably it's better if we wait until 2-3 weeks before the major release because in other case 2 charts should be maintained :/

JorTurFer:
During the context of this task, who will do it have to take in consideration that the resources are pending to split. They'd be separated [here](https://github.com/kedacore/charts/issues/182), but we decided to wait until this task to avoid one breaking change

AshutoshNirkhe:
This will indeed be a nice update!

JorTurFer:
I think that it's the moment to address this issue. I created it almost 2 years ago and we still don't plan a major KEDA release. 
As we have recently introduced the "big changes" in the architecture and we don't plan to introduce new big changes soon, I'd prepare the v3 of the chart, refactoring the whole chart, solving all the gaps and duplications that we currently have.

Currently, we have a chaotic chart with a lot of parameters without any sense (at least for me)
WDYT @kedacore/keda-helm-contributors ?

PR #239, opened 2022-01-27
feat: Provide parameters per container for resources & security context

### LINKED ISSUE #232: Can't have separate values like resources, securityContext for keda-operator and keda-operator-metrics-apiserver containers
## Expected Behavior
We should be able to define different values for parameters like `securityContext`, `resources `etc for `keda-operator` and `keda-operator-metrics-apiserver` containers.

Instead define hierarchy like,
```
.Values.operator.securityContext
.Values.operator-metrics-apiserver.securityContext
```

## Actual Behavior
Parameters like securityContext, resources etc are defined at root level i.e. .values.securityContext , they are applied to both containers keda-operator and keda-operator-metrics-apiserver containers.

None

## Specifications

  - **KEDA Version:** *v2.5.1*
  - **Platform & Version:** *Please elaborate*
  - **Kubernetes Version:** *v1.21.7*
  - **Scaler(s):** *Please elaborate*

AshutoshNirkhe:
Separate parameters like resources, securityContext for keda-operator and metrics-apiserver, fixes #232

Signed-off-by: AshutoshNirkhe <ashutosh.nirkhe@gmail.com>

### Checklist

Fixes #232


keda/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Event-based autoscaler for workloads on Kubernetes
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.1
+version: 2.5.2
```
tomkerkhove, line 7: Can you revert this please?

keda/templates/12-keda-deployment.yaml
```diff
@@ -40,11 +40,11 @@ spec:
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
       securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+        {{- toYaml .Values.podSecurityContext.operator | nindent 8 }}
```
tomkerkhove, line 43: Can you make it backwards compatible please? So that it uses .Values.podSecurityContext when available, otherwise .Values.podSecurityContext.operator when specified

tomkerkhove (changes requested):

keda/templates/12-keda-deployment.yaml
JorTurFer: +1

keda/Chart.yaml
AshutoshNirkhe: reverted

keda/templates/12-keda-deployment.yaml
AshutoshNirkhe: Sure, updated. Please check again.

JorTurFer: The solution looks nice but I think is not enough, What about not exiting `.Values.podSecurityContext.operator` but existing `.Values.podSecurityContext.metricServer`

tomkerkhove: THat's true, however, we use an empty object by default so think that should be fine (but to be verified)

AshutoshNirkhe: May be we can do something like,

image

But I am not sure if we are over-complicating the things by making this backward compatible ?

keda/README.md
```diff
@@ -84,7 +84,7 @@ their default values.
 | `logging.operator.format`                                  | Logging format for KEDA Operator. Allowed values are 'console' & 'json'. | `console`                                        |
 | `logging.operator.timeFormat`                              | Logging time format for KEDA Operator. Allowed values are 'epoch', 'millis', 'nano', or 'iso8601'. | `epoch` |
 | `logging.metricServer.level`                               | Logging level for Metrics Server.Policy to use to pull Docker images. Allowed values are '0' for info, '4' for debug, or an integer value greater than 0, specified as string | `0` |
```
tomkerkhove, line 86: Can you add the new configuration optoins here please?

keda/values.yaml
```diff
@@ -89,15 +89,21 @@ logging:
...
-  # fsGroup: 2000
+  # operator:
+  #   fsGroup: 2000
+  # metricServer:
+  #   fsGroup: 2000
 
 securityContext: {}
-  # capabilities:
-  #   drop:
-  #   - ALL
```
tomkerkhove, line 97: Please add this to the docs of the new flavors

AshutoshNirkhe: I will, but I thought they don't make much sense with runAsNonRoot: true

keda/templates/12-keda-deployment.yaml
AshutoshNirkhe: Ideally we could have made 2 top level sections for operator and metricServer and put parameters specific to those under respective section and common parameters off course live at top level, something like -

image

But I guess that's out of scope for this PR :)

keda/README.md
AshutoshNirkhe: added

keda/values.yaml
AshutoshNirkhe: Added

keda/README.md
```diff
@@ -84,9 +84,15 @@ their default values.
...
-| `securityContext`                                          | Security context of the containers within pod. ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)) | `{}` |
-| `podSecurityContext`                                       | Pod security context of the pod ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)) | `{}` |
-| `resources`                                                | Manage resource request & limits of KEDA workload ([docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)) | `{}` |
+| `securityContext`                                          | Security context of the containers (legacy option) ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)) | `{}` |
```
tomkerkhove, line 87:
```suggestion
| `securityContext`                                          | Security context for all containers ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)) | `{}` |
```

keda/README.md
```diff
@@ -84,9 +84,15 @@ their default values.
...
+| `securityContext.operator`                                 | Security context of the operator container ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)) | `` |
+| `securityContext.metricServer`                             | Security context of the metricServer container ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container)) | `` |
+| `podSecurityContext`                                       | Pod security context of the pods (legacy option) ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)) | `{}` |
```
tomkerkhove, line 90:
```suggestion
| `podSecurityContext`                                       | Pod security context for all pods ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)) | `{}` |
```

tomkerkhove (commented):
Let's not make it legacy but a convenience approach for when operator & metric server config is the same

keda/templates/12-keda-deployment.yaml
tomkerkhove: That is a fully rework and should be with the new major version imo

AshutoshNirkhe: Right. I don't want to break existing stuff, lol. But that's more of a suggestion for overall refactoring in future. But what about https://github.com/kedacore/charts/pull/239#discussion_r793545246 ?

tomkerkhove: I don't think we need that, but thanks for asking!

AshutoshNirkhe: Ok cool

keda/README.md
AshutoshNirkhe: Sure, updated

keda/templates/12-keda-deployment.yaml
JorTurFer: In fact, [we discussed](https://github.com/kedacore/charts/issues/182) it and we postponed till next major release.

For the moment, I'd maintain the backward compatibility and keep the refactors for the future. Maybe achieving this is more complicated from the chart pov, but IMHO it's the best approach.

tomkerkhove: > The solution looks nice but I think is not enough, What about not exiting `.Values.podSecurityContext.operator` but existing `.Values.podSecurityContext.metricServer`?

Even then we should use the current approach IMO because it's not because they define something for the metric server that the operator should have something

AshutoshNirkhe: So just confirming if we are good with what we have in the end ?

tomkerkhove: > > The solution looks nice but I think is not enough, What about not exiting `.Values.podSecurityContext.operator` but existing `.Values.podSecurityContext.metricServer`?
> 
> Even then we should use the current approach IMO because it's not because they define something for the metric server that the operator should have something

@JorTurFer Thoughts on this? Because with your current approach you enforce people to provide operator info when metric server info is specified.

djsly:
Thanks a lot @AshutoshNirkhe  for this work

keda/templates/12-keda-deployment.yaml
JorTurFer: We shouldn't assume that there is not a special user with super edge case, but for me it's enough if we continue as it is right now, adding a note explaining the limitation. Probably I'm overthinking the things, if we have any use case, we can do it and release a new chart version

tomkerkhove: > We shouldn't assume that there is not a special user with super edge case

But what is the super edge case? I'm not sure I'm getting the scenario you are concerned about?

What I believe we should provide is (similar for resources):
- `podSecurityContext.metricServer`: Applies pod security context to metrics server
- `podSecurityContext.operator`: Applies pod security context to operator
- `podSecurityContext`: Applied to both

Container specific variable takes priority over `podSecurityContext`. If they specify `podSecurityContext.operator` but not `podSecurityContext.metricServer` then that's their concern, we should not block them because of it im.

tomkerkhove: What are you thoughts @zroubalik ?

JorTurFer: > Container specific variable takes priority over podSecurityContext. If they specify podSecurityContext.operator but not podSecurityContext.metricServer then that's their concern, we should not block them because of it imo.

Yes, that's what I wanted to express :)

tomkerkhove: Ok, in that case we can merge the PR I think - Do you agree?

JorTurFer: yes

JorTurFer (approved):

tomkerkhove (approved):

PR #245, opened 2022-02-07
Add extra container args config for both keda operator and metric server

marco2704:
Currently neither the KEDA operator nor the metric server accept extra arguments in the KEDA chart. However, in some cases we may need to set additional arguments like when [How do I run KEDA with readOnlyRootFilesystem=true ? as describe in the KEDA FAQs](https://keda.sh/docs/2.5/faq/) where [metric-server flags](https://github.com/kubernetes-sigs/metrics-server/blob/master/docs/command-line-flags.txt) are needed.

This change enables the user to be able to set additional container arguments when needed.

### Checklist

- [X] README is updated with new configuration values *(if applicable)*


tomkerkhove (approved):
Thanks a ton!

PR #257, opened 2022-03-23
aggregate rbac for crds to default roles

### LINKED ISSUE #256: Include RBAC permissions to custom resources in KEDA helm chart
Like in https://github.com/kedacore/keda/issues/2723 for manifests, it would be great to add the option to deploy aggregation ClusterRoles for the build-in view, edit and admin roles.

### Use-Case

We provide keda as a service for cluster users and need to add the CRD permissions to the built-in view, edit, and admin ClusterRoles through [ClusterRole aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).
We deploy a separate keda-config helm chart with just these ClusterRoles.

If this is something you want to add, I'll be happy to provide a PR for this.

TheKangaroo:
this change adds ClusterRoles to aggregate rbac for
clustertriggerauthentications, scaledjobs, scaledobjects and
triggerauthentications to the kubernetes default roles view, edit and
admin via aggregation.
These ClusterRoles are optional and disabled by default. Set
rbac.aggregateToDefaultRoles to enable.

Signed-off-by: Till Adam <lxmail@web.de>

### Checklist

Fixes #256

### DIFF
--- a/keda/README.md
+++ b/keda/README.md
@@ -93,6 +93,7 @@ their default values.
 | `podLabels.metricsAdapter`                                 | Pod labels for KEDA Metrics Adapter | `{}` |
 | `podLabels.webhooks`                                       | Pod labels for KEDA Admission webhooks | `{}` |
 | `rbac.create`                                              | Specifies whether RBAC should be used | `true`                                        |
+| `rbac.aggregateToDefaultRoles`                             | Specifies whether RBAC for CRDs should be [aggregated](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) to default roles (view, edit, admin) | `false`                                        |
 | `serviceAccount.create`                                    | Specifies whether a service account should be created       | `true`                                        |
 | `serviceAccount.name`                                      | The name of the service account to use. If not set and create is true, a name is generated.      | `keda-operator` |
 | `serviceAccount.automountServiceAccountToken`              | Specifies whether created service account should automount API-Credentials | `true` |
--- a/keda/templates/manager/clusterrole.yaml
+++ b/keda/templates/manager/clusterrole.yaml
@@ -128,4 +128,52 @@ rules:
   - triggerauthentications/status
   verbs:
   - '*'
+{{- if .Values.rbac.aggregateToDefaultRoles }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: keda:edit
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    app.kubernetes.io/name: {{ .Values.operator.name }}
+    {{- include "keda.labels" . | indent 4 }}
+rules:
+- apiGroups:
+  - keda.sh
+  resources:
+  - clustertriggerauthentications
+  - scaledjobs
+  - scaledobjects
+  - triggerauthentications
+  verbs:
+  - create
+  - delete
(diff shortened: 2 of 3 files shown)


keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.17.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.6.2
+version: 2.6.3
```
JorTurFer, line 11: Please, revert this change because we bump the version during the chart release

JorTurFer (changes requested):
LGTM!
Thanks for this improvement! 
Only one small nit

JorTurFer:
@kedacore/keda-helm-contributors

TheKangaroo:
Ah thank you @JorTurFer, I reverted the version change.

tomkerkhove (approved):
LGTM, but be careful. We depend on what KEDA core generates as well so we might want to do it there as well.

TheKangaroo:
@tomkerkhove I'm not sure how keda core and helm chart plays together here, but if a PR in keda core is necessary, could you point me to the right direction?

tomkerkhove:
KEDA core generates YALM files to deploy but let's ask @zroubalik if this can be included first or if we should maintain it seperately

zroubalik:
Yeah, this should be changed in the core first. rbac is generated with kubebuilder from annotation on controllers, for example here: https://github.com/kedacore/keda/blob/9dbde78e238591870018a4eb9e2425b412ad26eb/controllers/keda/scaledobject_controller.go#L55-L63

tomkerkhove:
Are you up for contributing this in KEDA core as well @TheKangaroo?

TheKangaroo:
@tomkerkhove, sorry, this got lost on my side.
I had a look at @zroubalik's example and the kubebuilder docs, but I'm not sure how I can add aggregation roles via marker comments (i.e. add 'rbac.authorization.k8s.io/aggregate-to-xxxxx: "true"' labels to the generated clusterroles).

Or is ist just a matter of adding these clusterroles without any aggreation or bindings to the kedacore code? In this case, wouldn't it be better to drop the use of aggreation in this PR as well and just create clusterroles and clusterrolebindings instead?
Sorry if these are silly questions, I never worked with kubebuilder before.

tomkerkhove:
I will leave that up to @zroubalik to help though!

No worries, we are here to help you!

JorTurFer (approved):

JorTurFer:
We should give some attention to this @kedacore/keda-helm-contributors

TheKangaroo:
I had a brief chat with @zroubalik via slack about the kubebuilder part, but we didn't manage to find time to do this yet.
We can give this another shot if one of you can help me with some hints on how to do this.

JorTurFer:
We are waiting because kubebuiler doesn't support this and we generate the rbac files using it, but considering that the changes are in metadata section, I guess that can make then here for helm only. Wdyt @zroubalik @TheKangaroo ?

TheKangaroo:
I just rebased the PR.
We switched to the AKS managed keda in the meantime, so we do not use this chart anymore, but I still think this is useful for others as well.

u0078867:
Any news? Also interested here ...

zroubalik:
> As this is a role on top of other roles for giving access to KEDA resources as part of other cluster existing roles, I'd merge it without any change in core. @tomkerkhove @zroubalik ?

agree!

zroubalik (approved):
LGTM, thanks for the contribution!

PR #258, opened 2022-03-24
AutomountServiceAccountToken property implemented

ph311o:
Added new property "automountServiceAccountToken" in ServiceAccounts and Deployments necessary for restricted environments.

### Checklist

Fixes #

### DIFF
--- a/keda/templates/12-keda-deployment.yaml
+++ b/keda/templates/12-keda-deployment.yaml
@@ -39,6 +39,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      automountServiceAccountToken: true
       securityContext:
         {{- if .Values.podSecurityContext.operator }}
         {{- toYaml .Values.podSecurityContext.operator | nindent 8 }}
--- a/keda/templates/22-metrics-deployment.yaml
+++ b/keda/templates/22-metrics-deployment.yaml
@@ -42,6 +42,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      automountServiceAccountToken: true
       securityContext:
         {{- if .Values.podSecurityContext.metricServer }}
         {{- toYaml .Values.podSecurityContext.metricServer | nindent 8 }}
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -54,6 +54,8 @@ serviceAccount:
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name: keda-operator
+  # Specifies whether a service account should automount API-Credentials
+  automountServiceAccountToken: true
   # Annotations to add to the service account
   annotations: {}
 
--- a/keda/README.md
+++ b/keda/README.md
@@ -76,6 +76,7 @@ their default values.
 | `rbac.create`                                              | Specifies whether RBAC should be used | `true`                                        |
 | `serviceAccount.create`                                    | Specifies whether a service account should be created       | `true`                                        |
 | `serviceAccount.name`                                      | The name of the service account to use. If not set and create is true, a name is generated.      | `keda-operator` |
+| `serviceAccount.automountServiceAccountToken`              | Specifies whether created service account should automount API-Credentials | `true` |
 | `serviceAccount.annotations`                               | Annotations to add to the service account | `{}` |
 | `podIdentity.activeDirectory.identity`                     | Identity in Azure Active Directory to use for Azure pod identity | `` |
 | `grpcTLSCertsSecret`                                       | Name of the secret that will be mounted to the /grpccerts path on the Pod to communicate over TLS with external scaler(s) (recommended).  | ``|
--- a/keda/templates/01-serviceaccount.yaml
+++ b/keda/templates/01-serviceaccount.yaml
@@ -11,4 +11,5 @@ metadata:
   {{- end }}
   name: {{ .Values.serviceAccount.name }}
   namespace: {{ .Release.Namespace }}
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
 {{- end -}}


keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.17.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.6.2
+version: 2.6.3
```
tomkerkhove, line 11: Can you revert this please? We typically bump when we do a release

keda/templates/12-keda-deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
```
tomkerkhove, line 42: Let's use this approach instead:
```suggestion
      automountServiceAccountToken: {{ .Values.serviceAccount.automount.name }}
```

keda/templates/22-metrics-deployment.yaml
```diff
@@ -42,6 +42,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
```
tomkerkhove, line 45: Let's use this approach instead:
```suggestion
      automountServiceAccountToken: {{ .Values.serviceAccount.automount.name }}
```

keda/values.yaml
```diff
@@ -48,12 +48,17 @@ podDisruptionBudget: {}
 rbac:
   create: true
 
+# Specifies that both deployments (Keda/Metrics) should automount API-Credentials
```
tomkerkhove, line 51:
```suggestion
# Specifies that both deployments (KEDA/Metrics) should automount API-Credentials
```

tomkerkhove (changes requested):
Did a quick review, thanks for the change!

@zroubalik Do you want to have these changes in our generated YAML files?

keda/Chart.yaml
ph311o: Done.

keda/templates/12-keda-deployment.yaml
ph311o: @tomkerkhove Thank you for suggested change. Unfortunately this will not work as autmountServiceAccountToken is a boolean field. Because of security restrictions we need to set autmountServiceAccountToken in serviceAccount to false and in deployments to true.

keda/templates/22-metrics-deployment.yaml
ph311o: @tomkerkhove Thank you for suggested change. Unfortunately this will not work as autmountServiceAccountToken is a boolean field. Because of security restrictions we need to set autmountServiceAccountToken in serviceAccount to false and in deployments to true.

keda/values.yaml
ph311o: Done

keda/templates/12-keda-deployment.yaml
tomkerkhove: Let's go with this then?
```suggestion
      automountServiceAccountToken: {{ .Values.serviceAccount.automount.enabled }}
```

keda/templates/22-metrics-deployment.yaml
tomkerkhove: Let's go with this then?
```suggestion
      automountServiceAccountToken: {{ .Values.serviceAccount.automount.enabled }}
```

keda/templates/12-keda-deployment.yaml
tomkerkhove: Oh I get it. So it's not related to the service account but rather the deployment instead?

What about this then?
```suggestion
      automountServiceAccountToken: {{ .Values.deployment.automount.enabled }}
```

tomkerkhove: That's OK for me, thanks! I'm not an expert in this area, does `automountServiceAccountToken` have requirements or involve any risk to break existing installations?

ph311o: I have set serviceAccount and deployments to default behaviour of k8s so there is no risk for breaking existing installations. Additionally the needed opt-in for mounting API-token is added to both keda and metrics deployment. Please review. Thank you for your support.

JorTurFer:
Hey,
I'm looking for information about this feature and I can't find any clear use case (or most probably, I'm not understanding them). 
Could you explain your use case?

ph311o:
@JorTurFer Hi. Our UseCase is as follow:

In general Kubernetes automounts service account credentials which allows any compromised pod to run API commands against the cluster. We are restricted to stop this behaviour by specifying automountServiceAccountToken: false in ServiceAccount. This is a global rule so we cannot whitelist the helm chart.

As Keda and metrics deployments/pods need API access we need to opt in for these two with automountServiceAccountToken: true in both deployment.yamls

When you set both ServiceAccount and Deployments automountServiceAccountToken: true it's the same like not specifying at all so I am not sure if it makes sense to specify it in the installation.yaml. We need it in Helm Chart as said to disable default ServiceAccount behaviour and ensure functionality. 

Also see:

- https://github.com/kubernetes/kubernetes/issues/57601
- https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Grundschutz/IT-GS-Kompendium_Einzel_PDFs_2022/06_APP_Anwendungen/APP_4_4_Kubernetes_Edition_2022.pdf;jsessionid=831EB9783FD3996CD80511A21925E8EE.internet482?__blob=publicationFile&v=3#download=1

JorTurFer:
I got the reasons (I don't understand the document in German but I guess that it's an official regulation). 
My doubt is, how will KEDA work without its service account? I mean, KEDA needs its service account with the proper given permissions or basically KEDA doesn't work. 
I understand that common pods without any interaction with the API don't neet that service account, but KEDA needs it to work, Am I missing something? Are you enabling/injecting it in any other way?

ph311o:
@JorTurFer Do you need more informations from my side or do we wait for more opinions by other maintainers? In general you can disable it in ServiceAccount and enable it in Deployments hardcoded. But then everybody who would use the Keda ServiceAccount perhaps for other deployments could not longer access API. This was the reason why I decided to implement a switch in values.yaml for ServiceAccount.

zroubalik:
I think it is okay to have an optional switch, ie. default to the existing behavior.

ph311o:
@JorTurFer Yes absolutely it keeps current behaviour. Would be great if anybody could merge it for next release. Thank you.

keda/templates/22-metrics-deployment.yaml
```diff
@@ -42,6 +42,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      automountServiceAccountToken: true
```
tomkerkhove, line 45: Can you move this to a value with default property `true` instead please?

keda/templates/12-keda-deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      automountServiceAccountToken: true
```
tomkerkhove, line 42: Can you move this to a value with default property `true` instead please?

tomkerkhove (changes requested):

keda/templates/22-metrics-deployment.yaml
ph311o: Yes I can do that. I am not sure if this makes sense as you need that to set to true for having API Access. If you would set it to false KEDA would not work anymore. Please confirm if I should change it.

keda/templates/12-keda-deployment.yaml
ph311o: Yes I can do that. I am not sure if this makes sense as you need that to set to true for having API Access. If you would set it to false KEDA would not work anymore. Please confirm if I should change it.

tomkerkhove: With that you mean if `serviceAccount.automountServiceAccountToken` flag is on true? If so, why not re-use that same value then?

ph311o: Let me explain:
Every ServiceAccount generates an API Token regardless of what automountServiceAccountToken is set to (true, false or even not set)

Default behaviour when not specified automountServiceAccountToken in ServiceAccount and Deployment -> Pods using this ServiceAccount can access the API-Token.

ServiceAccount.automountServiceAccountToken: false -> All pods using this ServiceAccount have no access to API-Token anymore (but the ServiceAccount still have an API-Token stored internally)

Deployment can opt-in to regain access to API-Token:
Deployment.automountServiceAccountToken: true -> pods regain access to token

Does a switch make sense for ServiceAccount.automountServiceAccountToken?
Yes because e.g. we have a security policy that we can only deploy ServiceAccounts with ServiceAccount.automountServiceAccountToken: false.
So we have to define Deployment.automountServiceAccountToken: true in every Deployment which needs API-Access.

If somebody would use the ServiceAccount for own custom Deployments without specifying Deployment.automountServiceAccountToken they need to set ServiceAccount.automountServiceAccountToken: true. (Which would be the default switch value... So they do not recognize a change)

Does a switch make sense for Deployment.automountServiceAccountToken?
No. Because KEDA needs API-Access. A switch which could possible set Deployment.automountServiceAccountToken: false will break the API-Access for KEDA-Pods. (Regardless of ServiceAccount.automountServiceAccountToken)

Why even specifying Deployment.automountServiceAccountToken?
Because when ServiceAccount.automountServiceAccountToken: false and not specifying Deployment.automountServiceAccountToken the API-Access is gone.

Why not re-use the same value?
Because when setting ServiceAccount.automountServiceAccountToken: false then reusing value would result in Deployment.automountServiceAccountToken: false which prevents API-Access for pods.

tomkerkhove: Thanks for explaining!

tomkerkhove (approved):

tomkerkhove:
Would love to have a sign-off of @zroubalik or @JorTurFer as well though

JorTurFer (approved):
F**k, I missed this PR between the notifications :( 
If the behavior doesn't change, I don't have any problem to merge it

tomkerkhove:
Thank you @ph311o and sorry for being slow on our end!

PR #290, opened 2022-07-26
chore(helm): remove metrics-apiserver from podAntiAffinity example

aslafy-z:
Remove `keda-operator-metrics-apiserver` pods from podAntiAffinity example. 
There is no reason for preventing a `keda-operator` pod from scheduling on the same node as `keda-operator-metrics-apiserver`.

### Checklist

### DIFF
--- a/keda/Chart.yaml
+++ b/keda/Chart.yaml
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.17.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.7.2
+version: 3.0.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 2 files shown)


JorTurFer:
Hey,
that value is used for both deployments, so basically the sample specify that every instance of KEDA (operator or metrics server) has to run on a node without other KEDA instance. I'd not remove any of them because otherwise, multiple metrics server could be stuck in the same node.
Users can define that value for their use cases

JorTurFer:
you are right, there isn't any problem with that. The problem is that atm that section is used in both deployments, so removing that rule you can have operator an metrics server in the same node, but also more than 1 operator or metric server so potentially you don't have HA even you have more than 1 instance.

aslafy-z:
I see your concern, I will try to address it by updating the chart to add specific affinities fields for the two deployment if it's ok to you.

aslafy-z:
@JorTurFer I updated the PR with a preview of what I'd like to achieve. Do you think this is acceptable? It follows a standard pattern (I seen it everywhere) where a component is a top level object. Let me know so I can proceed. Thank you

tomkerkhove:
What is the status on this PR?

aslafy-z:
I just closed it. It looks like the affinities issue is fixed.

PR #295, opened 2022-08-02
feat(core): Provide capability to control replica count of metric server

tomkerkhove:
Provide capability to control replica count of metric server to reduce the impact when the active instance fails and allow for fast failover.

### Checklist

Relates to https://github.com/kedacore/keda/discussions/3386#discussioncomment-3230220


JorTurFer (approved):

keda/README.md
```diff
@@ -63,8 +63,9 @@ their default values.
...
 | `operator.affinity`                                        | Affinity for pod scheduling ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes-using-node-affinity/)) for KEDA operator. Takes precedence over the `affinity` field | `{}` |
+| `metricsServer.replicaCount`                                      | Capability to configure the number of replicas for KEDA metric server.<br /><br />While you can run more replicas of our metric server, only one instance will used and serve traffic.<br /><br />You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.<br /><br /> Learn more in [our documentation](https://keda.sh/docs/latest/operate/cluster/#high-availability).| `1` |
```
JorTurFer, line 68: This is not totally true, use all or not depends on an parameter but AKS, EKS and GKE AFAIK only query to one all the time.
IDK if it makes sense to clarify this or it's too in depth.
I could look for the parameter name because I think I have it anywhere if you want to clarify it

tomkerkhove: Can you elaborate please? I don't get what you are trying to say?

zroubalik: I don't get it as well. The description is quite accurate.

zroubalik (approved):

keda/README.md
JorTurFer: I remember a test we did where all the requests were routed to the same metrics server, but you told me @zroubalik  that it's a configuration parameter that OpenShift allows to override and you can also modify on bare metal clusters

JorTurFer: I can't figure out when we talked about this, so nvm

amirschw:
I guess this section of the docs also needs an update: https://keda.sh/docs/latest/operate/cluster/#high-availability

tomkerkhove:
Not really because the guidance still applies; but we just want to open it up for those who want to deviate from it

tomkerkhove:
Otherwise we'll have a lot of support cases for something that is a limitation

Issue #300, opened 2022-08-11
Chart 2.8.0 bundles outdated CRDs 


pinkfloydx33:
The CRDs in the 2.8.0 chart are not the actual CRDs for 2.8.0. For example the `spec.podIdentity.identityId` field is missing on `[Cluster]TriggerAuthentication`. 

I've double-checked that the CRDs inside the [keda-2.8.0.tgz](https://kedacore.github.io/charts/keda-2.8.0.tgz) file are not the latest ones

This is making migration a bit difficult. We installed Keda (originally) with `crds.install=true`. We have already upgraded to 2.8.0 but were unable to deploy any new trigger authentication resources due to bad CRDs. I patched the CRDs manually so I could deploy them, but they were destroyed on a follow-up `helm upgrade` used to adjust some other values.

## Specifications

  - **KEDA Version:** 2.8.0

tomkerkhove:
That's my fault, I will update them today - Sorry about that :/

PR #302, opened 2022-08-12
fix: Use CRDs from KEDA v2.8 & ship v2.8.1

tomkerkhove:
Use CRDs from KEDA v2.8 & ship v2.8.1

### Checklist

Fixes https://github.com/kedacore/charts/issues/300


JorTurFer (approved):
Update please the Readme with the versions please (it still says v2.7.1)

JorTurFer (approved):
Should we overwrite the previous chart too with this fix?
This is enough to me but maybe we can fix previous chart version

tomkerkhove:
No because then we'll have a shitshow because 2.8.0 actually is 2 versions: The one where people pulled it yesterday and what people will pull today. Releases should be immutable imo.

PR #307, opened 2022-08-19
feat: add seccompProfile

### LINKED ISSUE #301: metricServer can set readOnlyRootFilesystem=true
The keda [README](https://github.com/kedacore/charts/tree/main/keda) in this repo states that metricServer _cannot_ set readOnlyRootFilesystem=true

> Metrics server needs to write the self-signed cert so it's not possible set this
> readOnlyRootFilesystem: true

but the [FAQ](https://keda.sh/docs/2.8/faq/) says otherwise and provides instructions how to do it when using the default generated cert:

_Originally posted by @joebowbeer in https://github.com/kedacore/charts/pull/267#discussion_r943861229_

```yaml
    readOnlyRootFilesystem: true
  volumes:
    metricsApiServer:
      extraVolumes:
      - name: keda-volume
      extraVolumeMounts:
      - name: keda-volume
        mountPath: /apiserver.local.config/certificates/
```

Please add this secure snippet to the secure-by-default values, or add a reference to the FAQ in the comment.

### LINKED ISSUE #306: Add seccompProfile.type RuntimeDefault to KEDA deployments for secure-by-default
Add seccompProfile.type RuntimeDefault to Deployments as per PSS/restricted spec.

### Use-Case

This would bring the default manifests into compliance with PSS/restricted policies.

The policies currently fail because the containers in the two Deployments are missing an explicit seccompProfile type.

```bash
wget https://raw.githubusercontent.com/kyverno/policies/main/pod-security/restricted/restrict-seccomp-strict/restrict-seccomp-strict.yaml
kyverno apply restrict-seccomp-strict.yaml -r \
  <(helm template keda --repo https://kedacore.github.io/charts -n keda)
```

### Specification

https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

NOTE: seccompProfile is only supported in Kubernetes 1.19+

joebowbeer:
Add/update securityContext configs:
* seccompProfile.type RuntimeDefault is needed for compliance with restricted policies in Kubernetes 1.19+
* Updated comment in metrics Deployment and README regarding readOnlyRootFilesystem

### Checklist

Fixes #306 
Fixes #301

### DIFF
--- a/keda/README.md
+++ b/keda/README.md
@@ -172,13 +172,19 @@ securityContext:
       - ALL
     allowPrivilegeEscalation: false
     readOnlyRootFilesystem: true
+    ## Add explicit seccompProfile type for Kubernetes 1.19+
+    # seccompProfile:
+    #   type: RuntimeDefault
   metricServer:
     capabilities:
       drop:
       - ALL
     allowPrivilegeEscalation: false
-    ## Metrics server needs to write the self-signed cert so it's not possible set this
+    ## Metrics server needs to write the self-signed cert. See FAQ for discussion of options.
     # readOnlyRootFilesystem: true
+    ## Add explicit seccompProfile type for Kubernetes 1.19+
+    # seccompProfile:
+    #   type: RuntimeDefault
 
 podSecurityContext:
   operator:
--- a/keda/README.md
+++ b/keda/README.md
@@ -172,9 +172,8 @@ securityContext:
       - ALL
     allowPrivilegeEscalation: false
     readOnlyRootFilesystem: true
-    ## Add explicit seccompProfile type for Kubernetes 1.19+
-    # seccompProfile:
-    #   type: RuntimeDefault
+    seccompProfile:
+      type: RuntimeDefault
   metricServer:
     capabilities:
       drop:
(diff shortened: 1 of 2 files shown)


joebowbeer:
NOTE: seccompProfile is only supported in Kubernetes 1.19+

This is causing the 1.17 and 1.18 tests to fail.

tomkerkhove:
It is best to use Helm operators to check if the feature is available to avoid this, can you take a look at this? This is typically done through the API version

joebowbeer:
@tomkerkhove wrote:
> It is best to use Helm operators to check if the feature is available to avoid this, can you take a look at this? This is typically done through the API version

I think the options are:
1. Add code to the templates to conditionally remove these elements from the replacement
2. Comment-out these elements in values until 1.19 is the minimum supported version

tomkerkhove:
> Add code to the templates to conditionally remove these elements from the replacement

I think that's the best approach

joebowbeer:
@tomkerkhove  PTAL

Since seccompProfile requires Kubernetes 1.19+ (otherwise validation fails), I think this is the best that _should_ be done until the minimum version is raised from 1.17 to 1.19.

While it should be possible to finagle the values here, it is not possible in keda/config (https://github.com/kedacore/keda/pull/3562) and I think it's better to keep the two in sync.

tomkerkhove:
Let's make sure the change is being applied because nobody will notice the comment.

As per https://github.com/kedacore/governance/blob/main/SUPPORT.md#kubernetes-support I'll enforce Kubernetes v1.20.

joebowbeer:
@tomkerkhove PTAL

I restored the seccompProfile changes.

Will you bump the k8s minVersion to 1.19?

tomkerkhove (approved):
LGTM, thanks!

PR #312, opened 2022-08-29
topology: add support for pod topology constraints to operator

erikkn:
This PR adds support for  `Pod Topology Constraints` for the `keda-operator` Pod.

### Checklist

Fixes #

### DIFF
--- a/keda/README.md
+++ b/keda/README.md
@@ -106,6 +106,7 @@ their default values.
 | `resources.metricServer`                                   | Manage resource request & limits of KEDA metrics apiserver pod ([docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)) | `` |
 | `nodeSelector`                                             | Node selector for pod scheduling ([docs](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/)) | `{}` |
 | `tolerations`                                              | Tolerations for pod scheduling ([docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)) | `{}` |
+| topologySpreadConstraints | object | `{}` | Pod Topology Constraints https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/ |
 | `affinity`                                                 | Affinity for pod scheduling ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes-using-node-affinity/)) for both KEDA operator and Metrics API Server | `{}` |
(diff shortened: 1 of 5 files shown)


tomkerkhove:
Any reason why we are not adding this for metric server as well?

@zroubalik Do you want to have this in CORE as well?

erikkn:
@tomkerkhove correct me if I am wrong, but based on the [documentation](https://keda.sh/docs/2.1/operate/cluster/#high-availability), my understanding is that currently, only one replica of the keda-metrics-server is supported; in that case Pod Topology Constraints for this Pod is a bit useless, or do you disagree entirely :)?

tomkerkhove:
That is true, but same goes for the operator :) But it's not 100% the same limitation so it's fine, thanks for the PR!

erikkn:
The operator supports multiple (>=2) replicas that also does leader-election.

Please let me know if you want me to add support for the `keda-metrics-server` as well, very happy to add it.

tomkerkhove:
It's OK, we can leave metric server out

JorTurFer:
[We can have more than 1 instance in the metrics server](https://github.com/kedacore/charts/pull/295). Even the traffic could be routed to a single instance (upstream limitation), recently we added support to specify more than 1 instance in metrics server.
I said because maybe we should extend this PR to include metrics server

erikkn:
> [We can have more than 1 instance in the metrics server](https://github.com/kedacore/charts/pull/295). Even the traffic could be routed to a single instance (upstream limitation), recently we added support to specify more than 1 instance in metrics server. I said because maybe we should extend this PR to include metrics server

Alrighty, I will add support for that later tonight/tomorrow, thanks!

erikkn:
@JorTurFer I added support for the `metricsServer` as well, you mind reviewing this PR :)?

keda/README.md
```diff
@@ -106,6 +106,7 @@ their default values.
 | `resources.metricServer`                                   | Manage resource request & limits of KEDA metrics apiserver pod ([docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)) | `` |
 | `nodeSelector`                                             | Node selector for pod scheduling ([docs](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/)) | `{}` |
 | `tolerations`                                              | Tolerations for pod scheduling ([docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)) | `{}` |
+| topologySpreadConstraints | object | `{}` | Pod Topology Constraints https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/ |
```
tomkerkhove, line 109: This is not correct as the correct fields are `topologySpreadConstraints.operator` & `topologySpreadConstraints.metricsServer`

tomkerkhove (approved):
LGTM, but we need to update the README

PR #316, opened 2022-09-20
metrics: move prometheusrules under metricServer

erikkn:
Hi, folks!

I don't mean to be pedantic, but the `keda-operator-metrics-apiserver` app is exposing these (https://keda.sh/docs/2.8/operate/prometheus/) metrics. 
At the moment we create Prometheus rules under the `prometheus.operator` directive which is subsequently used in `keda/templates/15-keda-prometheusrules.yaml`. The name of the Prometheus rule is also `{{ .Values.operator.name }}`.

This all is a bit confusing to me and frankly, doesn't make sense, since it belongs to the `metricServer` section.

Would like to her you guys opinion on this :)

### Checklist

- [ x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [ x] README is updated with new configuration values *(if applicable)*

Fixes #


erikkn:
@tomkerkhove decided to just tag you on this since you also reviewed my last PR. I would love to hear your opinion on this; thanks!

keda/values.yaml
```diff
@@ -298,3 +286,15 @@ prometheus:
         #   expr: sum by ( scaledObject , scaler) (rate(keda_metrics_adapter_scaler_errors[2m]))  > 0
         #   for: 2m
         #   labels:
+  operator:
```
tomkerkhove, line 289: Dont think you wanted to use `operator`?

tomkerkhove (commented):
The problem here is taht this change will break a lot of end-users. We'll need to be backwards compatible here (even if it's confusing).

Side question - @zroubalik Are we planning on exposing Prometheus metrics on the operator?

erikkn:
I do recognize that this is a breaking change, you want to support both for now?

tomkerkhove:
I think it's best to do, yes

tomkerkhove:
We actually already have metrics on the operator so we should make sure both are still accessible in a backwards compatible way - https://github.com/kedacore/keda/issues/3586

So instead of removing the operator pieces, I'd just keep them in addition to your new additions.

@JorTurFer as FYI

v-shenoy:
I think having metrics on both pods is the way, as it makes sense to record certain metrics on particular pods. So, having changes both would be great.

tomkerkhove:
What is the status on this PR? This might be a bit outdated due to our recent Prometheus changes

PR #319, opened 2022-09-30
Added support for AWS IRSA in helm chart

### LINKED ISSUE #233: Add helm chart option for EKS IRSA account
Add helm chart support for the [required AWS configuration](https://keda.sh/docs/2.2/troubleshooting/#why-does-keda-operator-error-with-nocredentialproviders) on the ServiceAccount.  

### Use-Case

We're using IRSA for auth inside AWS EKS.  Therefore, the helm chart should offer an out-of-the-box ability to specify the annotation `eks.amazonaws.com/role-arn` if using IRSA.

### Specification

- [ ] Add an optional field for IRSA role arn to the `values.yaml` file

StevenJDH:
Signed-off-by: Steven Jenkins De Haro 20492442+StevenJDH@users.noreply.github.com

Adds support for AWS IAM Roles for Service Accounts (IRSA) in the helm chart to the KEDA service account. For example, if `podIdentity.irsa.enabled` is set `true` along with the needed config, the below will be rendered:

```yaml
apiVersion: v1
kind: ServiceAccount
...
  annotations:
    eks.amazonaws.com/audience: "sts.amazonaws.com"
    eks.amazonaws.com/role-arn: "arn:aws:iam::000000000000:role/example-irsa-role"
    eks.amazonaws.com/sts-regional-endpoints: "true"
    eks.amazonaws.com/token-expiration: "3600"
...
```

### Checklist

- [ X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [ X] A PR is opened to update KEDA core ([repo](https://github.com/kedacore/keda)) *(if applicable, ie. when deployment manifests are modified)*
- [X ] README is updated with new configuration values *(if applicable)*

Fixes #233

### DIFF
--- a/keda/README.md
+++ b/keda/README.md
@@ -89,11 +89,11 @@ their default values.
 | `podIdentity.azureWorkload.enabled`                        | Specifies whether [Azure Workload Identity](https://azure.github.io/azure-workload-identity/) is to be enabled or not. ([docs](https://keda.sh/docs/concepts/authentication/#azure-workload-identity)) | `false` |
 | `podIdentity.azureWorkload.tenantId`                       | Id Azure Active Directory Tenant to use for authentication with for Azure Workload Identity. ([docs](https://keda.sh/docs/concepts/authentication/#azure-workload-identity)) | `` |
 | `podIdentity.azureWorkload.tokenExpiration`                | Duration in seconds to automatically expire tokens for the service account. ([docs](https://keda.sh/docs/concepts/authentication/#azure-workload-identity)) | `3600` |
-| `podIdentity.irsa.audience`                                | Sets the token audience for IRSA. | `sts.amazonaws.com` |
-| `podIdentity.irsa.enabled`                                 | Specifies whether [AWS IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) is to be enabled or not. | `false` |
-| `podIdentity.irsa.roleArn`                                 | ARN of an IRSA IAM role with a web identity provider to use for authentication via STS. | `` |
-| `podIdentity.irsa.stsRegionalEndpoints`                    | Sets the use of an STS regional endpoint instead of global. Recommended to use regional endpoint in almost all cases. | `true` |
-| `podIdentity.irsa.tokenExpiration`                         | Duration in seconds to automatically expire tokens for the service account. | `86400` |
+| `podIdentity.aws.irsa.audience`                            | Sets the token audience for IRSA. | `sts.amazonaws.com` |
+| `podIdentity.aws.irsa.enabled`                             | Specifies whether [AWS IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) is to be enabled or not. | `false` |
+| `podIdentity.aws.irsa.roleArn`                             | ARN of an IRSA IAM role with a web identity provider to use for authentication via STS. | `` |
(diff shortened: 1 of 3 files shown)


keda/README.md
```diff
@@ -89,6 +89,11 @@ their default values.
...
 | `podIdentity.azureWorkload.tokenExpiration`                | Duration in seconds to automatically expire tokens for the service account. ([docs](https://keda.sh/docs/concepts/authentication/#azure-workload-identity)) | `3600` |
+| `podIdentity.irsa.audience`                                | Sets the token audience for IRSA. | `sts.amazonaws.com` |
+| `podIdentity.irsa.enabled`                                 | Specifies whether [AWS IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) is to be enabled or not. | `false` |
+| `podIdentity.irsa.roleArn`                                 | ARN of an IRSA IAM role with a web identity provider to use for authentication via STS. | `` |
+| `podIdentity.irsa.stsRegionalEndpoints`                    | Sets the use of an STS regional endpoint instead of global. Recommended to use regional endpoint in almost all cases. | `true` |
+| `podIdentity.irsa.tokenExpiration`                         | Duration in seconds to automatically expire tokens for the service account. | `86400` |
```
tomkerkhove, line 96: Let's rename these to clarify it for end-users who are not familiar with AWS
```suggestion
| `podIdentity.awsIrsa.audience`                                | Sets the token audience for IRSA. | `sts.amazonaws.com` |
| `podIdentity.awsIrsa.enabled`                                 | Specifies whether [AWS IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) is to be enabled or not. | `false` |
| `podIdentity.awsIrsa.roleArn`                                 | ARN of an IRSA IAM role with a web identity provider to use for authentication via STS. | `` |
| `podIdentity.awsIrsa.stsRegionalEndpoints`                    | Sets the use of an STS regional endpoint instead of global. Recommended to use regional endpoint in almost all cases. | `true` |
| `podIdentity.awsIrsa.tokenExpiration`                         | Duration in seconds to automatically expire tokens for the service account. | `86400` |
```

tomkerkhove (changes requested):
LGTM, but we'd need to rename the variables

keda/README.md
tomkerkhove: Maybe `podIdentity.aws.irsa.*` is even better in case AWS comes up with another provider (which we should have done for Azure as well)

tomkerkhove: > Sounds like a good idea. I think this future proof option is also a better approach. So the agreement is that I update the PR to reflect the `podIdentity.aws.irsa.*` pattern?

Correct!

> I will keep the focus of the changes on AWS for his PR, but perhaps in another PR we can align Azure Workload Identity to follow the same pattern, but this would be a breaking change.

Yes, we can open an issue for that to track it but can't change it for now

tomkerkhove (approved):
LGTM, thanks!

ivanmarquescepsa:
hi, i would need this feature, do you have an estimate of when will it be released?

Thanks!

hlastraspfx:
Same question: when will it be released? Any news?

JorTurFer:
Next KEDA release will be around Dec 8th, and it includes the chart. You can check the release dates [here](https://github.com/kedacore/keda/blob/main/ROADMAP.md)

PR #328, opened 2022-11-21
Add workload identity labels to keda pods

v-shenoy:
Signed-off-by: Vighnesh Shenoy <vshenoy@microsoft.com>

The `azure.workload.identity/use` label will now be required for pods as well (besides the service account).

https://github.com/Azure/azure-workload-identity/pull/629

### Checklist

Fixes #

### DIFF
--- a/keda/templates/12-keda-deployment.yaml
+++ b/keda/templates/12-keda-deployment.yaml
@@ -30,6 +30,9 @@ spec:
         {{- if .Values.podLabels.keda }}
         {{- toYaml .Values.podLabels.keda | nindent 8}}
         {{- end }}
+        {{- if .Values.podIdentity.azureWorkload.enabled }}
+        azure.workload.identity/use: {{ .Values.podIdentity.azureWorkload.enabled | quote }}
+        {{- end }}
       {{- if .Values.podAnnotations.keda }}
       annotations:
       {{- toYaml .Values.podAnnotations.keda | nindent 8}}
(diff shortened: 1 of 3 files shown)


JorTurFer:
Do we need both annotations since now? I mean, are both required or it's one or the other?

v-shenoy:
> Do we need both annotations since now? I mean, are both required or it's one or the other?



> Do we need both annotations since now? I mean, are both required or it's one or the other?

https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html

From what I get from this page is that both are required.

keda/templates/12-keda-deployment.yaml
```diff
@@ -30,6 +30,9 @@ spec:
         {{- if .Values.podLabels.keda }}
         {{- toYaml .Values.podLabels.keda | nindent 8}}
         {{- end }}
+        {{- if .Values.podIdentity.azureWorkload.enabled }}
+        azure.workload.identity/use: {{ .Values.podIdentity.azureWorkload.enabled | quote }}
```
tomkerkhove, line 34: Nit - Either we hard-code it if we use an IF, or we always add the label based on the property.

My preference would be this:
```suggestion
        azure.workload.identity/use: "true"
```

keda/templates/22-metrics-deployment.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   labels:
     app: {{ .Values.operator.name }}-metrics-apiserver
     app.kubernetes.io/name: {{ .Values.operator.name }}-metrics-apiserver
+    {{- if .Values.podIdentity.azureWorkload.enabled }}
+    azure.workload.identity/use: {{ .Values.podIdentity.azureWorkload.enabled | quote }}
```
tomkerkhove, line 10: See other comment

tomkerkhove (approved):

JorTurFer (approved):

PR #335, opened 2022-11-26
Support disable keep http connection alive

penghuazhou:
Support disable keep http connection alive

### Checklist

Relates to https://github.com/kedacore/keda/pull/3878
Relates to https://github.com/kedacore/keda/issues/3874

### DIFF
--- a/keda/templates/12-keda-deployment.yaml
+++ b/keda/templates/12-keda-deployment.yaml
@@ -106,6 +106,10 @@ spec:
             - name: KEDA_HTTP_DEFAULT_TIMEOUT
               value: {{ .Values.http.timeout | quote }}
             {{- end }}
+            {{- if ( not .Values.http.keepAlive.enabled ) }}
+            - name: KEDA_HTTP_DISABLE_KEEP_ALIVE
+              value: "true"
+            {{- end }}
             {{- if .Values.env }}
             {{- toYaml .Values.env | nindent 12 -}}
             {{- end }}
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -131,7 +131,7 @@ podIdentity:
       # Set to the value of the ARN of an IAM role with a web identity provider.
       # This will be set as an annotation on the KEDA service account.
       roleArn: ""
-      # Sets the use of an STS regional endpoint instead of global. 
+      # Sets the use of an STS regional endpoint instead of global.
       # Recommended to use regional endpoint in almost all cases.
       # This will be set as an annotation on the KEDA service account.
       stsRegionalEndpoints: "true"
@@ -259,6 +259,8 @@ priorityClassName: ""
 ## reasonable default
 http:
   timeout: 3000
+  keepAlive:
+    enabled: true
 
 ## Extra KEDA Operator and Metrics Adapter container arguments
 extraArgs:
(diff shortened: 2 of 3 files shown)


tomkerkhove:
@penghuazhou Can you please fill in the PR template?

keda/templates/12-keda-deployment.yaml
```diff
@@ -100,6 +100,10 @@ spec:
             - name: KEDA_HTTP_DEFAULT_TIMEOUT
               value: {{ .Values.http.timeout | quote }}
             {{- end }}
+            {{- if .Values.http.disableKeepAlive }}
```
tomkerkhove, line 103: Let's use this approach which follows our typical way of configuring flags
```suggestion
            {{- if .Values.http.keepAlive.enabled }}
```

tomkerkhove (changes requested):

keda/templates/12-keda-deployment.yaml
JorTurFer: Remember during the change that this requires changes in docs and code too ;)

penghuazhou: @tomkerkhove @JorTurFer , as http transports is "DisableKeepAlives", so i transform the value in helm, please review again, thanks.
image

penghuazhou:
> @penghuazhou Can you please fill in the PR template?

done

keda/templates/12-keda-deployment.yaml
```diff
@@ -100,6 +100,10 @@ spec:
             - name: KEDA_HTTP_DEFAULT_TIMEOUT
               value: {{ .Values.http.timeout | quote }}
             {{- end }}
+            {{- if .Values.http.keepAlive.enabled }}
+            - name: KEDA_HTTP_DISABLE_KEEP_ALIVE
+              value: {{ .Values.http.keepAlive.enabled }}
+            {{- end }}
```
JorTurFer, line 106: We can't do this as the current code has the opposite logic, when `.Values.http.keepAlive.enabled` is true, we disable it in the code.
I know that the code is already merged, so maybe we can handle this here adding some logic, for example:
```suggestion
            {{- if not( .Values.http.keepAlive.enabled) }}
            - name: KEDA_HTTP_DISABLE_KEEP_ALIVE
              value: "true"
            {{- end }}
```

And the same for the other manifest. WDYT?

JorTurFer (changes requested):

keda/templates/12-keda-deployment.yaml
penghuazhou: @JorTurFer, done

keda/values.yaml
```diff
@@ -254,6 +254,8 @@ priorityClassName: ""
 ## reasonable default
 http:
   timeout: 3000
+  keepAlive:
+    enabled: false
```
JorTurFer, line 258: This should be `true` as default to maintaining current behaviour. When this is `true`, we don't add the env variable, so we don't disable the keep alive

```suggestion
    enabled: true
```

JorTurFer (approved):

JorTurFer:
Can we merge this before the release @tomkerkhove ?

tomkerkhove (approved):

Issue #342, opened 2022-12-09
Unknown flag `--metrics-port` when enabling prometheus on the operator


nextrevision:
## Expected Behavior

Enabling scraping on the prometheus operator should create a PodMonitor.

```yaml
prometheus:
  operator:
    enabled: true
    podMonitor:
      enabled: true
```

## Actual Behavior

The pod fails to start with the following error:

```
Usage of /keda:
      --disable-compression                   Disable response compression for k8s restAPI in client-go.  (default true)
      --health-probe-bind-address string      The address the probe endpoint binds to. (default ":8081")
      --kube-api-burst int                    Set the burst for throttling requests sent to the apiserver (default 30)
      --kube-api-qps float32                  Set the QPS rate for throttling requests sent to the apiserver (default 20)
      --kubeconfig string                     Paths to a kubeconfig. Only required if out-of-cluster.
      --leader-elect                          Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.
      --metrics-bind-address string           The address the metric endpoint binds to. (default ":8080")
      --metrics-service-bind-address string   The address the gRPRC Metrics Service endpoint binds to. (default ":9666")
      --zap-devel                             Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
      --zap-encoder encoder                   Zap log encoding (one of 'json' or 'console') (default )
      --zap-log-level level                   Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity (default )
      --zap-stacktrace-level level            Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
      --zap-time-encoding time-encoding       Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'. (default )
unknown flag: --metrics-port
```

The metrics port in the template: https://github.com/kedacore/charts/blob/main/keda/templates/12-keda-deployment.yaml#L75

## Steps to Reproduce the Problem

  1. Enable prometheus scraping on the operator via values.
  2. Deploy latest chart.

## Specifications

  - **KEDA Version:** *2.9.0*
  - **Platform & Version:** *EKS*
  - **Kubernetes Version:** *1.23*
[...]

PR #343, opened 2022-12-11
fix(keda): Remove unknown flag when enabling prometheus on the operator

JorTurFer:
Signed-off-by: Jorge Turrado <jorge_turrado@hotmail.es>

This PR removes the old parameter `--metrics-port` and modifies the CI check covering the monitoring as part of the check. Also replaces the tag for CI to enforce main, in order to detect this errors faster (currently we check with latest released version, which could be outdated for testing the chart because latest release doesn't contain incoming changes)

### Checklist

Fixes https://github.com/kedacore/charts/issues/342

### DIFF
--- a/.github/workflows/ci-core.yml
+++ b/.github/workflows/ci-core.yml
@@ -82,6 +82,11 @@ jobs:
     - name: Generate values
       run: |
         cat <<EOF > test-values.yaml
+        image:
+          keda:
+            tag: main
+          metricsApiServer:   
+            tag: main
         podIdentity:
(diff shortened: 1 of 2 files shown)


tomkerkhove (approved):

tomkerkhove:
So our current chart is broken then and we need a patch version?

JorTurFer:
For users with prometheus operator enabled, the chart is broken yes, we should release the new chart version IMO

tomkerkhove:
Let's do it!

Issue #359, opened 2023-01-09
Provide support for scheduling highly available across node per zone


tomkerkhove:
Provide support for scheduling highly available across node per zone to avoid outages due to pod scheduling on same node.

PR #360, opened 2023-01-09
feat(http-add-on): Provide support for scheduling highly available across node per zone

tomkerkhove:
Provide support for scheduling highly available across node per zone

### Checklist

Fixes https://github.com/kedacore/charts/issues/359

### DIFF
--- a/http-add-on/README.md
+++ b/http-add-on/README.md
@@ -88,8 +88,10 @@ their default values.
 | `images.scaler`                                            | Image name for the scaler image component | `ghcr.io/kedacore/http-add-on-scaler:latest` |
 | `images.kubeRbacProxy.name`                                | Image name for the Kube RBAC Proxy image component | `gcr.io/kubebuilder/kube-rbac-proxy` |
 | `images.kubeRbacProxy.tag`                                 | Image tag for the Kube RBAC Proxy image component | `v0.5.0` |
-| `additionalLabels`                                         | Additional labels to be applied to installed resources. Note that not all resources will receive these labels. | Nothing |
 | `crds.install`                                             | Whether to install the `HTTPScaledObject` [`CustomResourceDefinition`](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) | `true` |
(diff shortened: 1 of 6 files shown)


JorTurFer (commented):
Why not just `podTopologySpread`? I mean, does `highAvailability` give value here? For HA scenarios there are other things required like multi replicas, pod disruption budgets, etc...
I'd put `podTopologySpread`  inside each component because they can be in different pools for different reasons, IMO each component should be configurable

tomkerkhove:
> Why not just `podTopologySpread`? I mean, does `highAvailability` give value here? For HA scenarios there are other things required like multi replicas, pod disruption budgets, etc...

Because I'll add more things under that umbrella for future features, such as pod disruption budgets :) We tend to have too many top level things and should provide more structure.

> I'd put `podTopologySpread` inside each component because they can be in different pools for different reasons, IMO each component should be configurable

The chart is having too many knobs while they will generally be enabled for the same thing. One will not enable HA for interceptor; but not for the others. That makes no sense to me.

JorTurFer:
> The chart is having too many knobs while they will generally be enabled for the same thing. One will not enable HA for interceptor; but not for the others. That makes no sense to me.

Are all the components ready for HA? I think the scaler can't have more than 1 instance. This could be more confusing if users think that they have an HA configuration but we have a single instance. TBH, I prefer to expose all the parameters but not doing "black magic" under the hood, we can expose the topology key, with an explanation as comments in value file or having an example's folder rather than hidding the config

tomkerkhove:
I think your approach is coming from a technical perspective and mine is around making it simple for end-users.

HA is more than running multiple instances and I think we should start with a general configuration, and give more granular knobs if people want them. That way, the configuration is simple until we need to go more advanced.

tomkerkhove:
Let's see what @zroubalik thinks and I'll adapt accordingly

JorTurFer:
@zroubalik , what do you think?

tomkerkhove:
@zroubalik Thoughts?

JorTurFer:
Kindly reminder @zroubalik , wdyt?

tomkerkhove:
It's a nob they can turn on so that we schedule things highly available; but it's fine - Let's abandon this PR

PR #375, opened 2023-01-26
fix: ensure chart works on k8s 1.17

JorTurFer:
Signed-off-by: Jorge Turrado <jorge.turrado@scrm.lidl>

### Checklist

Fixes https://github.com/kedacore/keda/issues/4126

### DIFF
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -153,8 +153,9 @@ securityContext:
       - ALL
     allowPrivilegeEscalation: false
     readOnlyRootFilesystem: true
-    seccompProfile:
-      type: RuntimeDefault
+    # k8s >= 1.19
+    # seccompProfile: 
+    #   type: RuntimeDefault
   metricServer:
     capabilities:
       drop:
@@ -162,8 +163,9 @@ securityContext:
     allowPrivilegeEscalation: false
     ## Metrics server needs to write the self-signed cert. See FAQ for discussion of options.
     # readOnlyRootFilesystem: true
-    seccompProfile:
-      type: RuntimeDefault
+    # k8s >= 1.19
+    # seccompProfile:
+    #   type: RuntimeDefault
 
 podSecurityContext:
   operator:
--- a/.github/workflows/ci-core.yml
+++ b/.github/workflows/ci-core.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         enableAzureWorkloadIdentity: [false, true]
-        kubernetesVersion: [v1.24, v1.23, v1.22, v1.21, v1.20]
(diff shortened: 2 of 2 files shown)


JorTurFer:
I have tried to add k8s 1.17 to the CI using helm but I haven't been able. I have tried deploying the chart locally with a 1.17 cluster

keda/values.yaml
```diff
@@ -153,17 +153,19 @@ securityContext:
       - ALL
     allowPrivilegeEscalation: false
     readOnlyRootFilesystem: true
-    seccompProfile:
-      type: RuntimeDefault
+    # k8s >= 1.19
```
tomkerkhove, line 156: Let's not punish those using Kubernetes 1.20 or above; we should use the .Capabilties functionality to see if seccompProfile is supported.

https://github.com/Azure/api-management-self-hosted-gateway/blob/94217377b23c9c0571b2dc90a1dfcced3ca8aa73/helm-charts/azure-api-management-gateway/templates/ingress-stable.yaml#L1

tomkerkhove (changes requested):

keda/values.yaml
JorTurFer: Can I get `seccompProfile` support from capabilities? I thought it gives only API support but not at this level

JorTurFer: I guess that I can check k8s version and ignore the value in case of not supported, but that's worse for me because I'd be changing given user fields. Honestly, I prefer to comment the field (in v2.8 only) but I can't add a note saying that the value will be ignored in k8s <= 1.18 and continue with the previous default value

tomkerkhove: Why not check if Kubernetes version is v1.19 or above and only then apply seccompProfile? That is more user friendly than assume people will see the note and manually apply it

JorTurFer: Because people can expect it applied in clusters <1.19 as they set the values and we would magically remove it.
Also this change won't go to main, it's only for v2.8 release (which in theory should have a short life)
But I can do your proposal if you prefer

tomkerkhove: > Because people can expect it applied in clusters <1.19 as they set the values and we would magically remove it.

That is fine; it is skipping a configuration because it is not supported on their cluster which is safe to do. If we want to be transparent; we can add the same if in NOTES.txt and mention it was skipped

> Also this change won't go to main, it's only for v2.8 release (which in theory should have a short life)

That's fine; but we still need to do proper fixes.

> But I can do your proposal if you prefer

I think this should be done, but let's ask @zroubalik for a 3rd opinion.

tomkerkhove (approved):

PR #377, opened 2023-01-27
Release: HTTP Add-on v0.4.0

JorTurFer:
Signed-off-by: Jorge Turrado <jorge.turrado@scrm.lidl>

### Checklist


http-add-on/values.yaml
```diff
@@ -132,7 +132,7 @@ images:
   # stable release in the official images, `canary` is 
   # the build for the latest commit to the `main` branch,
   # and you can target any other commit with `sha-<GIT_SHA[0:7]>`
-  tag: 0.3.0
+  tag:
```
tomkerkhove, line 135: Just to be sure; if it's not specifies it relies on chart.appversion?

tomkerkhove (approved):

http-add-on/values.yaml
JorTurFer: yeah https://github.com/kedacore/charts/blob/main/http-add-on/templates/deployment-interceptor.yaml#L47

JorTurFer: I think that having only once (in chart.appversion) is better

PR #379, opened 2023-01-28
fix: Remove duplicate key in http-addon rbac

abatilo:
Noticed this while trying to use the kustomize helm chart inflator

Signed-off-by: Aaron Batilo <AaronBatilo@gmail.com>


JorTurFer (approved):

JorTurFer:
Thanks for the fix!

PR #388, opened 2023-02-03
chore: add support for configuring KEDA_HTTP_MIN_TLS_VERSION

JorTurFer:
Signed-off-by: Jorge Turrado <jorge_turrado@hotmail.es>

### Checklist

Related to https://github.com/kedacore/keda/issues/4193

### DIFF
--- a/keda/templates/22-metrics-deployment.yaml
+++ b/keda/templates/22-metrics-deployment.yaml
@@ -86,10 +86,10 @@ spec:
           env:
             - name: WATCH_NAMESPACE
               value: {{ .Values.watchNamespace | quote }}
-            {{- if .Values.http.timeout }}
             - name: KEDA_HTTP_DEFAULT_TIMEOUT
               value: {{ .Values.http.timeout | quote }}
-            {{- end }}
+            - name: KEDA_HTTP_MIN_TLS_VERSION
+              value: {{ .Values.http.minTlsVersion }}
             {{- if ( not .Values.http.keepAlive.enabled ) }}
             - name: KEDA_HTTP_DISABLE_KEEP_ALIVE
               value: "true"
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -325,6 +325,7 @@ http:
   timeout: 3000
   keepAlive:
     enabled: true
+  minTlsVersion: TLS12
 
 ## Extra KEDA Operator and Metrics Adapter container arguments
 extraArgs:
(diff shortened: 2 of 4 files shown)


keda/templates/22-metrics-deployment.yaml
```diff
@@ -86,10 +86,10 @@ spec:
           env:
             - name: WATCH_NAMESPACE
               value: {{ .Values.watchNamespace | quote }}
-            {{- if .Values.http.timeout }}
             - name: KEDA_HTTP_DEFAULT_TIMEOUT
               value: {{ .Values.http.timeout | quote }}
-            {{- end }}
+            - name: KEDA_HTTP_MIN_TLS_VERSION
```
tomkerkhove, line 91: What about our interceptor/webhooks?

JorTurFer: Webhooks don't have any HttpClient (created by us). I was thinking about supporting thisor not in the metrics server as the metric querying is deprecated there, and I added it for security reasons more than for giving options

tomkerkhove: Ok; fair enough

keda/values.yaml
```diff
@@ -325,6 +325,7 @@ http:
   timeout: 3000
   keepAlive:
     enabled: true
+  minTlsVersion: TLS12
```
tomkerkhove, line 328: Can you update the readme with the new setting please?

tomkerkhove (approved):

keda/templates/22-metrics-deployment.yaml
```diff
@@ -86,10 +86,10 @@ spec:
           env:
             - name: WATCH_NAMESPACE
               value: {{ .Values.watchNamespace | quote }}
-            {{- if .Values.http.timeout }}
```
zroubalik, line 89: this is not needed anymore?

JorTurFer: I think so because [values file defines a default value](https://github.com/kedacore/charts/blob/main/keda/values.yaml#L325). I mean, that if is always true

keda/values.yaml
JorTurFer: Done!, I didn't see the comment yesterday sorry

keda/templates/22-metrics-deployment.yaml
JorTurFer: If you prefer, I can add it again, but I removed to keep templates cleanest as possible

zroubalik (approved):

Issue #394, opened 2023-02-17
Extra objects manifests


fboula:
Enable the possibility to add some extra manifest to the chart by passing a yaml list of k8s manifests. 

This would allow to create any objects you may require to make your keda install to work. This is extremely useful and easy to maintain, an example can be seen in [argo workflow chart](https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/values.yaml#L562)

### Use-Case

- Create `ClusterTriggerAuthentication`
- Create secrets required for a `ClusterTriggerAuthentication` 
- Any other resources that would make sense to deploy along the keda chart

If this gets approved, I can work on the pull-request to implement it.

mkilchhofer:
It is really not good practice to deploy CRDs and resources of that CRDs (= an instance) inside one helm release.

See the comment I made in your PR:
- https://github.com/kedacore/charts/pull/395#discussion_r1190861060

PR #395, opened 2023-02-17
feat: Support extra kubernetes manifests

fboula:
Added support for extra kubernetes manifest creation. Fixes https://github.com/kedacore/charts/issues/394

### Checklist

- [x] README is updated with new configuration values


keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.23.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.9.4
```
JorTurFer, line 11: please, don't bump this version. we do it during release process

JorTurFer (commented):
looking good! Could you add any extra object to the [test case](https://github.com/kedacore/charts/blob/main/.github/workflows/ci-core.yml#L85)? 🙏 

What do you think about this feature @kedacore/keda-helm-contributors ?

keda/Chart.yaml
fboula: Sure

keda/README.md
```diff
@@ -214,7 +214,7 @@ their default values.
 | `certificates.certManager.enabled`                         | Enables Cert-manager for certificate management | `false` |
 | `certificates.certManager.generateCA`                      | Generates a self-signed CA with Cert-manager | `true` |
 | `certificates.certManager.caSecretName`                    | Secret name where the CA is stored (generatedby cert-manager or user given) | `kedaorg-ca` |
-
+| `extraObjects`                                             | Array of extra K8s manifests to deploy | `[]`|
```
tomkerkhove, line 217: Proposal
```suggestion
| `manifests.extra`                                             | Array of extra K8s manifests to deploy | `[]`|
```

keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.23.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.10.0
+version: 2.9.4
```
tomkerkhove, line 11: Can you please revert this?

JorTurFer:
Could you check the e2e tests too? It seems that the extra manifests are applied before the others and that's why they are failing (because ClusterTriggerAuthentication isn't registered yet)

.github/workflows/ci-core.yml
```diff
@@ -120,6 +120,15 @@ jobs:
...
             generateCA: true
+        extraObjects:
+          - apiVersion: keda.k8s.io/v1alpha1
+            kind: ClusterTriggerAuthentication
+            metadata:
+              name: aws-credentials
+              namespace: keda
+            spec:
+              podIdentity:
+                provider: aws-eks
```
JorTurFer, line 131: I think that here you have a race condition because `ClusterTriggerAuthentication` isn't created before trying to render the extra object. Maybe using any kind of hook to delay the process you can achieve this

keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.24.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.10.1
```
JorTurFer, line 11: Please, don't bump the version, we do it during releases

mkilchhofer:
I also need this feature to be able to deploy NetworkPolicies :-) Is there something I could assist?

JorTurFer:
@fboula , any update ?

.github/workflows/ci-core.yml
```diff
@@ -120,6 +120,26 @@ jobs:
...
+        extraObjects:
+          - apiVersion: v1
+            kind: Secret
+            metadata:
+              name: aws-credentials
+              namespace: keda
+            type: Opaque
+            data:
+              AWS_ACCESS_KEY_ID: RkFLRUFXU0FDQ0VTU0tFWUlE
+              AWS_SECRET_ACCESS_KEY: RkFLRUFXU1NFQ1JFVEFDQ0VTU0tFWQ==
```
mkilchhofer, line 132: I'd change it to something different than a Secret. Having a Secret in an example will encourage users to do this and having plain secrets in a section what is human readable is not nice.

Maybe something like we use in argo-helm would fit better:
https://github.com/argoproj/argo-helm/blob/5873a1bc331e37c9500e4337d2ad2deca2fd8c80/charts/argo-cd/values.yaml#L535-L560

Or if you want to do an example with aws-creds I'd use at least 
- `kind: SealedSecret` (https://github.com/bitnami-labs/sealed-secrets)
- `kind: ExternalSecret` (https://external-secrets.io/)

keda/templates/02-crd-clustertriggerauthentications.yaml
```diff
@@ -4,6 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.9.0
+    helm.sh/hook: crd-install
```
mkilchhofer, line 7: This is not supported anymore with helm v3. This was the mechanism of helm v2

fboula: Hey I just checked that in helm v3 docs, they encourage to use a dedicated `crds` repository instead. I am willing to do it, does keda need to maintain the helm v2 compatibility at all ?

.github/workflows/ci-core.yml
fboula: Gotcha

keda/templates/02-crd-clustertriggerauthentications.yaml
mkilchhofer: I'd not change anything as its perfectly fine like it is for the KEDA chart. You should just not make an example with a resource of CRD kind the chart itself deploys.

If crds are moved outside the `/templates` folder into `/crds` the chart itself cannot update the CRDs if needed. Therefore I encourage you to not move anything regarding the CRDs.

fboula:
@JorTurFer I added a post-install hook on the test case. It looks better, thanks for the hint.

keda/templates/02-crd-clustertriggerauthentications.yaml
zroubalik: @mkilchhofer agree, good point.

fboula:
@JorTurFer any updates ?

JorTurFer (approved):

JorTurFer:
Thanks for the improvement ❤️

PR #398, opened 2023-03-01
feat(http-addon): narrow permissions

t0rr3sp3dr0:
Create new Roles and ServiceAccounts for the Interceptor and for the Scaler with granular permissions given to them. The Operator Role was changed to narrow its scope and remove write access to resources not owned by it.

See kedacore/http-add-on#613

### Checklist

Fixes kedacore/http-add-on#612


http-add-on/templates/rbac-scaler.yml
```diff
@@ -0,0 +1,92 @@
...
+  - list
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - watch
```
JorTurFer, line 51: Does the scaler need these permissions?

JorTurFer (approved):
LGTM, only one question inline

http-add-on/templates/rbac-scaler.yml
t0rr3sp3dr0: Right now it does to check the number of ready replicas of a service. I’m planning to change this on a future patch and make it rely only on the endpoints information instead of getting that from the deployment status.

PR #402, opened 2023-03-09
chore: add minutes and seconds to cert-manager certificates

JorTurFer:
In some scenarios using gitops (like using ArgoCD), the current certificate spec isn't enough because minutes and seconds are missing. As the desired manifest (without min a sec) doesn't match with the live manifest (fulfilled with min and sec), ArgoCD shows an OutOfSync warning. This is skippable by config, but if the chart uses the full values, ignoring isn't needed
image

### Checklist

### DIFF
--- a/keda/templates/41-cert-manager-self-ca.yaml
+++ b/keda/templates/41-cert-manager-self-ca.yaml
@@ -11,8 +11,8 @@ spec:
   privateKey:
     algorithm: RSA
     size: 2048
-  duration: 8760h # 1 year
-  renewBefore: 5840h # 8 months
+  duration: 8760h0m0s # 1 year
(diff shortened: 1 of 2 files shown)


tomkerkhove (approved):

tomkerkhove:
Isn't the diff the other way around, based on the code change?

JorTurFer:
> Isn't the diff supposed to be the other way around, based on the code change?

Yes, but when the server adds extra info that can trigger the diff. In this case we are setting `5840h`, which is valid, and the server is adding the minutes and the seconds because even though a value without them is valid, Kubernetes stores the object with them. 
ArgoCD checks the desired manifest (without min a secs) with the live manifest patched by k8s (with min and secs) and that's why the diff is triggered because there is a change indeed.

Issue #410, opened 2023-03-16
Bad indentation in 22-metrics-deployment.yaml


joebowbeer:
The indentation for extraVolumes and extraVolumeMounts in 22-metrics-deployment.yaml is incorrect.

## Expected Behavior

Can render the v2.10.x chart with an extra volume and extra volume mount, as per instructions for setting `readOnlyRootFilesystem` in [FAQ](https://keda.sh/docs/2.10/faq/).

```bash
helm template keda kedacore/keda --namespace keda \
  --set 'volumes.metricsApiServer.extraVolumes[0].name=keda-volume' \
  --set 'volumes.metricsApiServer.extraVolumeMounts[0].name=keda-volume' \
  --set 'volumes.metricsApiServer.extraVolumeMounts[0].mountPath=/apiserver.local.config/certificates/' \
  --set 'securityContext.metricServer.readOnlyRootFilesystem=true'
```

## Actual Behavior

```
Error: YAML parse error on keda/templates/22-metrics-deployment.yaml: error converting YAML to JSON: yaml: line 95: did not find expected key
```

## Steps to Reproduce the Problem

  1. Run command above (from FAQ)

## Specifications

  - **KEDA Version:** 2.10.1
  - **Platform & Version:** N/A
  - **Kubernetes Version:** N/A
  - **Scaler(s):** N/A

JorTurFer:
Hey,
Nice catch! With KEDA v2.10 you don't need to mount any extra volume for it because thanks to new certificates, the metrics server doesn't create certs (they are created by the operator or externally and mounted in the metrics server)
You can set `readOnlyRootFilesystem: true` directly and it should work (and thanks because I have noticed the missing changes)

PR #412, opened 2023-03-16
chore: enable `readOnlyRootFilesystem: true` for metrics server

JorTurFer:
Thanks to the cert-management, KEDA mestrics server always receives the certificates as a mounted path, so we can set `readOnlyRootFilesystem: true` as default

### Checklist

Related to https://github.com/kedacore/charts/issues/410

### DIFF
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -226,8 +226,7 @@ securityContext:
       drop:
       - ALL
     allowPrivilegeEscalation: false
-    ## Metrics server needs to write the self-signed cert. See FAQ for discussion of options.
-    # readOnlyRootFilesystem: true
+    readOnlyRootFilesystem: true
     seccompProfile:
       type: RuntimeDefault
   webhooks:


keda/values.yaml
```diff
@@ -226,8 +226,7 @@ securityContext:
       drop:
       - ALL
     allowPrivilegeEscalation: false
-    ## Metrics server needs to write the self-signed cert. See FAQ for discussion of options.
-    # readOnlyRootFilesystem: true
+    readOnlyRootFilesystem: true
```
tomkerkhove, line 229: Is cert manager required feature or not? If not, LGTM

tomkerkhove (approved):

keda/values.yaml
JorTurFer: No it isn't. With or without cert manager, the cert is a dep that will be always mounted in the metrics server (generated by cert-manager or by keda-operator)

joebowbeer (commented):
I assume this is thanks to the read-only certs volume that is always added now.

Where does the [kedaorg-certs](https://github.com/kedacore/charts/blob/1ca4b046e312d1cc9bb805588055e6bec5b8d330/keda/values.yaml#L444) secret come from?

JorTurFer:
If you aren't using cert-manager, it's KEDA operator who creates and manages that certificate and the service patching. 
If you enable cert-manager, will be cert-manager who does those stuff.

This is a feature introduced in v2.10 to improve the security

JorTurFer:
You can get more information from docs https://keda.sh/docs/2.10/operate/security/#use-your-own-tls-certificates

PR #431, opened 2023-04-18
Add resource template to kube-rbac-proxy container in http-add-on helm chart

danielamar101:
The kube-rbac-proxy container in the operator deployment is missing the `resources` template. In environments where limits are required to be set on all deployments, this causes manual labor for k8s developers. 

I would update the README to reflect this change, however, the README is already essentially wrong. 
for example [The http-add-on README](https://github.com/kedacore/charts/blob/main/http-add-on/README.md#L138)
states that there is a configuration for `interceptor.resources.limits.cpu`, while in the deployment template it uses a top level resource declaration `{{- toYaml .Values.resources | nindent 12 }}`. I could instead close this MR and create another one to instead fix the templates to reflect the README.
### Checklist

### DIFF
--- a/http-add-on/templates/deployment-operator.yaml
+++ b/http-add-on/templates/deployment-operator.yaml
@@ -39,6 +39,8 @@ spec:
         - --logtostderr=true
         - --v=10
         image: "{{ .Values.images.kubeRbacProxy.name }}:{{ .Values.images.kubeRbacProxy.tag }}"
+        resources:
+          {{- toYaml .Values.resources | nindent 12 }}
         name: kube-rbac-proxy
       - args:
         - --metrics-addr=127.0.0.1:8080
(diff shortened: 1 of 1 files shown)


http-add-on/templates/deployment-operator.yaml
```diff
@@ -39,6 +39,8 @@ spec:
         - --logtostderr=true
         - --v=10
         image: "{{ .Values.images.kubeRbacProxy.name }}:{{ .Values.images.kubeRbacProxy.tag }}"
+        resources:
+          {{- toYaml .Values.resources | nindent 12 }}
```
JorTurFer, line 43: Nice catch but I think that we can hard-code the values because it's only for the metrics proxy, otherwise we will request too much resources for it

danielamar101: good point. Just changed it. Does this look better?

JorTurFer (approved):

http-add-on/templates/deployment-operator.yaml
JorTurFer: yes it does, maybe we need to increase or allow updates in the future, but as starting point is good enough

PR #439, opened 2023-05-09
adding correct annotations for svcms and fixing sa

tkennes:

Fixes #
- Wrong annotations for Prometheus service monitors
- Bug in indentation of the keda-operator serviceaccount


keda/templates/17-keda-podmonitor.yaml
```diff
@@ -3,8 +3,10 @@ apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
   name: {{ .Values.operator.name }}
+  {{- with .Values.prometheus.operator.podMonitor.annotations }}
```
tomkerkhove, line 6: This is going to break `additionalAnnotations`

keda/templates/17-keda-servicemonitor.yaml
```diff
@@ -3,8 +3,10 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ .Values.operator.name }}
+  {{- with .Values.prometheus.operator.serviceMonitor.annotations }}
```
tomkerkhove, line 6: This is going to break `additionalAnnotations`

keda/templates/26-metrics-podmonitor.yaml
```diff
@@ -3,8 +3,10 @@ apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
   name: {{ .Values.operator.name }}-metrics-apiserver
+  {{- with .Values.prometheus.metricServer.podMonitor.annotations }}
```
tomkerkhove, line 6: This is going to break `additionalAnnotations`

keda/templates/27-metrics-servicemonitor.yaml
```diff
@@ -3,8 +3,10 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ .Values.operator.name }}-metrics-apiserver
+  {{- with .Values.prometheus.metricServer.serviceMonitor.annotations }}
```
tomkerkhove, line 6: This is going to break `additionalAnnotations`

keda/templates/33-webhooks-servicemonitor.yaml
```diff
@@ -4,8 +4,10 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ .Values.webhooks.name }}
+  {{- with .Values.prometheus.webhooks.serviceMonitor.annotations }}
```
tomkerkhove, line 7: This is going to break `additionalAnnotations`

tomkerkhove (changes requested):
Added some breaking changes, but would you mind also fixing DCO?

tkennes:
@tomkerkhove : I don't really understand the lack of flexible when it comes to setting annotations for specific resources, could you explain?

Also, would it not make more sense to opt for some sort of global + local mechanism to set labels? See also here for example: https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml#L3

tkennes:
@tomkerkhove : I don't really understand the lack of flexible when it comes to setting annotations for specific resources, could you explain?

Also, would it not make more sense to opt for some sort of global + local mechanism to set labels? See also here for example: https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml#L3

tkennes:
@tomkerkhove : I don't really understand the lack of flexible when it comes to setting annotations for specific resources, could you explain?

Also, would it not make more sense to opt for some sort of global + local mechanism to set labels? See also here for example: https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml#L3

tkennes:
@tomkerkhove , in that case, it depends a bit on the cluster whether annotations are heavily used for individual resources. As a rule of thumb, you would use labels for most things especially when it comes to managing infrastructure and tying certain resources together. E.g. a service binds to a pod based on the label, same for servicemonitors and podmonitors. This is also why you cannot really use labels if you intend to segment resources further, as you will have to be careful not to misuse any labels. Hence, you use annotations. I have seen companies use annotations to route alerts to the right teams for example.

Therefore, it might make sense to allow for individual annotations per resource rather than having only one global annotation. If you see benefit in the approach, I can set this up for all resources like for ArgoCD (https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml#L3), e.g., we merge global .Values.additionalAnnotations with local resource specific ones, like for example .Values.prometheus.metricServer.serviceMonitor.additionalAnnotations.

tkennes:
> Added some breaking changes, but would you mind also fixing DCO?
@tomkerkhove , this should be resolved by now right?

tomkerkhove:
This is OK for me, can you pull in the changes from main & document the new options please?

tomkerkhove:
We should document all options, even if they are well-known ones because people who are new to the cloud-native space would not know about them unfortunately.

Given https://github.com/kedacore/charts/issues/444 is being done, it's OK to just document them in the values file, thanks!

tomkerkhove (approved):
LGTM, thanks!

Would you mind fixing the merge conflicts please?

Issue #437, opened 2023-05-08
Helm chart not correctly rendering when using additional annotations and azure workload identity


tkennes:
## Expected Behavior
Helm template rendering

## Actual Behavior
Issue. Cannot convert YAML to JSON

## Steps to Reproduce the Problem

  1. Set podIdentity.azureWorkload.enabled: true
  2. Set additionalAnnotations: key: value
  3. Run a helm template dry-run

## Specifications

  - **KEDA Version:** *2.10.2*

## Fix
- Change indent to nindent here: https://github.com/kedacore/charts/blob/main/keda/templates/01-serviceaccount.yaml#L14, or remove the annotation on the service account altogether.

I tried to submit a pull request myself, but I'm not allowed to push a branch.

jared-schmidt-niceincontact:
I'm trying to use additionalAnnotations from the 2.9.4 chart which has nindent in it and it isn't working either.  There's something broken going on with the annotations.  I'm wondering if some neighboring field is gobbling whitespace incorrectly.

tomkerkhove:
Looks like the issue is probably here:
`  annotations:    keda: keda`

Are you willing to contribute a fix?

tkennes:
Exactly, that's why I proposed the nindent :).


Sure. But I'll provide a couple more fixes. The reason I have this weird annotation is because of issues on my servicemonitors, it's a bit wonky with resources in a permanent state of out-of-sync under ArgoCD management.

jared-schmidt-niceincontact:
I opened a PR (#441) that revamps annotations all over the helm chart so they will actually work correctly.  There is a problem where the '{}' gets included in the annotations even when you override the value with something else.  There's also an issue with empty annotations.  So, I modified that in all of the templates.

PR #441, opened 2023-05-15
Fixes labels and annotations

jared-schmidt-niceincontact:
Labels and annotations have been fixed throughout the templates so that they actually work.  The "annotation" key in the metadata is included or excluded depending on if it has content.  The "{}" in several of the annotation overrides in the values.yaml file was causing the "helm template" command to break when a value was defined in an external values.yaml file.  This bug has also been fixed.

### Checklist

- [X] README is updated with new configuration values *(if applicable)*

Fixes #
#437


keda/values.yaml
```diff
@@ -271,8 +270,8 @@ service:
   portHttpTarget: 8080
   portHttps: 443
   portHttpsTarget: 6443
-
-  annotations: {}
+  # annotations:
+  #   foo: bar
```
mkilchhofer, line 274: Don't comment-out the key.

keda/values.yaml
```diff
@@ -135,7 +133,8 @@ serviceAccount:
   # Specifies whether a service account should automount API-Credentials
   automountServiceAccountToken: true
   # Annotations to add to the service account
-  annotations: {}
+  # annotations:
+  #   foo: bar
```
mkilchhofer, line 137: Don't comment-out the key.

keda/values.yaml
```diff
@@ -105,14 +105,12 @@ podDisruptionBudget: {}
...
-  # foo: bar
+# additionalLabels:
+#   foo: bar
 
 # -- Custom annotations to add into metadata
-additionalAnnotations:
-  {}
-  # foo: bar
+# additionalAnnotations:
+#   foo: bar
```
mkilchhofer, line 113: Don't comment-out the key.

keda/values.yaml
```diff
@@ -105,14 +105,12 @@ podDisruptionBudget: {}
   #   maxUnavailable: 1
 
 # -- Custom labels to add into metadata
-additionalLabels:
-  {}
-  # foo: bar
+# additionalLabels:
+#   foo: bar
```
mkilchhofer, line 109: Don't comment-out the key.

keda/values.yaml
```diff
@@ -399,7 +398,7 @@ prometheus:
       targetPort:
       interval:
       scrapeTimeout:
-      relabellings: []
+      relabelings: []
```
mkilchhofer, line 401: This is kinda breaking. But if we accept this, this change needs to be reflected into the 3 servicemonitor.yaml files:
- templates/manager/servicemonitor.yaml
- templates/webhooks/servicemonitor.yaml
- templates/metrics-server/servicemonitor.yaml

keda/values.yaml
```diff
@@ -376,7 +375,7 @@ prometheus:
...
```
mkilchhofer, line 378: This is kinda breaking. But if we accept this, this change needs to be reflected into the 3 servicemonitor.yaml files:
- templates/manager/servicemonitor.yaml
- templates/webhooks/servicemonitor.yaml
- templates/metrics-server/servicemonitor.yaml

mkilchhofer (changes requested):
Additional if you change keys inside values.yaml you need to also adapt the README.md

keda/values.yaml
tomkerkhove: Agreed, we should not break this

keda/values.yaml
tomkerkhove: Agreed, we should not break this

tomkerkhove (changes requested):
LGTM, but I have the same comments such as @mkilchhofer. Please process them and also undo all changes in the values file wrt commenting things

PR #442, opened 2023-05-15
chore: Simplify cert-manager certs and support existingIssuer

mkilchhofer:
Improve 2 things related to cert-manager certificates:
- Simplify the cert-manager resources. 2 of 4 resources are not required and I never seen such setups in the community
- add support for an existing Issuer / ClusterIssuer.

Somehow related to #371 as this introduced the usage of cert-manager.

### Checklist

- [ ] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)

Fixes #


keda/templates/metrics-server/apiservice.yaml
```diff
@@ -3,11 +3,7 @@ kind: APIService
 metadata:
   annotations:
     {{- if .Values.certificates.certManager.enabled }}
-    {{- if .Values.certificates.certManager.generateCA }}
-    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Values.operator.name }}-ca
-    {{- else }}
-    cert-manager.io/inject-ca-from-secret: {{ .Release.Namespace }}/{{ .Values.certificates.certManager.caSecretName }}
-    {{- end }}
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Values.operator.name }}-tls-certificates
```
mkilchhofer, line 6: was configured like its documented inside the cert-manager docs.
I'd suggest to follow 1:1 the docs of cert-manager:
https://cert-manager.io/docs/concepts/ca-injector/#injecting-ca-data-from-a-certificate-resource

JorTurFer: What do you mean?

mkilchhofer: The annotation `cert-manager.io/inject-ca-from` has the value `namespace-name/certificate-name`

In the current implementation of the chart the value is `namespace-name/secret-name` which seems also working but it does not follow the documentation.

JorTurFer: you are right

JorTurFer:
I believe that we will agree with a [v3 of the chart](https://github.com/kedacore/charts/issues/182). If finally we continue with the new version, I think that it'll be the good moment to address this change too

mkilchhofer:
Rebased :-)

JorTurFer:
I still think that we should work on KEDA chart v3 and remove all this code that we think it's not okey. It brings us the flexibility to break everything. 
I like your approach of removing the intermediate issuer, but it's a breaking change for users who are using the secret with their own CA to generate the certificates...
@tomkerkhove @zroubalik , do we have consensus about working on a chart v3 from scratch (this will be a breaking change and also will detach versioning between core and chart)

zroubalik:
> I still think that we should work on KEDA chart v3 and remove all this code that we think it's not okey. It brings us the flexibility to break everything. I like your approach of removing the intermediate issuer, but it's a breaking change for users who are using the secret with their own CA to generate the certificates... @tomkerkhove @zroubalik , do we have consensus about working on a chart v3 from scratch (this will be a breaking change and also will detach versioning between core and chart)

I am all in for Charts simplification. I am not sure what we should do with the versioning though...mabye we can even start thinking about KEDA v3? ... there are several proposal that would be breaking and actually great https://github.com/kedacore/keda/issues/4771
But I don't know...

mkilchhofer:
Reusing an existing issuer is now possible via:
- https://github.com/kedacore/charts/pull/530

Closing my issue now 👍

PR #443, opened 2023-05-16
optionaly install apiservice

shcherbak:
Adding ability to optionally install APIService/v1beta1.external.metrics.k8s.io in case of more then one release of KEDA in the cluster

Even if two keda operators are defined with `operator-name` and `watch-namespace` values there is one more shared resource 
```
SharedResourceWarning
APIService/v1beta1.external.metrics.k8s.io is part of applications APP1 and APP2
```
there is permanent concurention  for this resource between two ArgoCD applications in one cluster

### DIFF
--- a/keda/Chart.yaml
+++ b/keda/Chart.yaml
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.24.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.10.2
+version: 2.10.3
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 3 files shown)


keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.24.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.10.2
+version: 2.10.3
```
JorTurFer, line 11: don't bump this version. We do it during the chart release process

JorTurFer (commented):
Are you using multiple metrics servers? This change will enable multiple KEDAs for ScaledJobs but for ScaledObjects, the HPA controller won't work for ScaledObjects managed by other KEDA installations

keda/Chart.yaml
shcherbak: @JorTurFer i have realized that two installations will not work because the metrics server, so I will use single keda setup
`name: v1beta1.external.metrics.k8s.io` should be defined only one time (

JorTurFer: so, should we continue with this PR or we can close it?

PR #452, opened 2023-05-23
fix(keda): APIService port comes from service.portHttps

hyperbolist:

### Checklist

### Fixes

- `APIService` now takes its `port` from `service.portHttps`.  Previously, the APIService's port was omitted in the manifest, which resulted in the port defaulting to 443, even if we actually set the port to something else via service.portHttps.  When this happens, you'll see its availability check as false with reason `ServicePortError` and a message complaining that the APIService is not available on the port specified in its manifest.


JorTurFer (changes requested):
nice catch with the APIService!
as we don't release all the changes on each commit, could you undo the other changes? (keeping only the change in the APIService manifest)

hyperbolist:
> as we don't release all the changes on each commit, could you undo the other changes?

Of course, reverted.

JorTurFer (approved):

JorTurFer:
Thanks for the improvement! :heart:

PR #454, opened 2023-06-01
fix(http-add-on): Fix resources variables for the http-add-on deploym…

jocelynthode:
Fixes the deployment templates to actually use the `resources` values defined in the `values.yaml` file. Currently they are ignored and no resources requests or limits are set for the `http-add-on`.

### Checklist

### DIFF
--- a/http-add-on/templates/deployment-interceptor.yaml
+++ b/http-add-on/templates/deployment-interceptor.yaml
@@ -82,7 +82,7 @@ spec:
         - containerPort: {{ .Values.interceptor.proxy.port }}
           name: inter-proxy
         resources:
-          {{- toYaml .Values.resources | nindent 12 }}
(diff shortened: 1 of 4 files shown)


JorTurFer (approved):
nice catch!

jocelynthode:
@JorTurFer Thanks for the quick review and merge! Do you know when I can expect this change to be live in the helm repo?

JorTurFer:
I reckon that we could do a helm release now, as this value is useful for user adoption. WDYT @tomkerkhove?

jocelynthode:
@JorTurFer @tomkerkhove: Hey, sorry to bother you about this, but I'd really like to get a release for this fix as we're in the process of adding requests/limits to everything in our cluster and `keda-http-add-on` is one of the last missing piece and using upstream instead of vendoring in the chart is always easier to maintain :)

PR #470, opened 2023-06-20
feat: Allow to change port for Webhooks and add hostnetwork

### LINKED ISSUE #468: Allow to set hostnetwork on Webhook
### Discussed in https://github.com/kedacore/charts/discussions/462

Allow to use Webhook with the argument `hostnetwork: true`.

### Use-Case

To address limitations faced on EKS with custom CNIs such as Calico, Weave, or Cilium, we require the ability to use webhooks with the hostnetwork: true argument. More details on this can be found in the following article: https://medium.com/@denisstortisilva/kubernetes-eks-calico-and-custom-admission-webhooks-a2956b49bd0d.

### Specification

- [ ] Enable the setting of hostnetwork on the webhook. You can find the relevant code in the Keda chart's values.yaml file here: https://github.com/kedacore/charts/blob/main/keda/values.yaml#L80.
- [ ] (Optional) Allow customization of the webhook port. This would require adding a new argument to the Keda, https://github.com/kedacore/keda/blob/0e7662d0fcb6d8084f2161ba70b9d457442f80f4/cmd/webhooks/main.go#L84.

mcantinqc:

### Checklist

Fixes #468

Need this change : https://github.com/kedacore/keda/pull/4712

### DIFF
--- a/keda/templates/webhooks/deployment.yaml
+++ b/keda/templates/webhooks/deployment.yaml
@@ -26,7 +26,7 @@ spec:
         app: {{ .Values.webhooks.name }}
         name: {{ .Values.webhooks.name }}
         app.kubernetes.io/name: {{ .Values.webhooks.name }}
-        {{- include "keda.labels" . | indent 8 }}       
+        {{- include "keda.labels" . | indent 8 }}
         {{- if .Values.podLabels.webhooks }}
         {{- toYaml .Values.podLabels.webhooks | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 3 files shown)


keda/templates/webhooks/validatingconfiguration.yaml
```diff
@@ -26,6 +26,7 @@ webhooks:
       name: {{ .Values.webhooks.name }}
       namespace: {{ .Release.Namespace }}
       path: /validate-keda-sh-v1alpha1-scaledobject
+      port: {{ .Values.webhooks.port | default 9443 }}
```
JorTurFer, line 29: This doesn't work because the port defined here is the service port (443) and the chart doesn't allow changing it. If you want to modify it, please add another parameter and update the service port as well.
Before the change, it worked because the default value is 443

JorTurFer: Just to clarify, this is the port used in the control plane to reach the hook (the service port), not the hook pod port itself (the target port)

zroubalik (approved):

PR #515, opened 2023-08-31
remove namespace from webhook

goenning:
Remove unnecessary `namespace` from metadata since ValidatingWebhookConfiguration is a cluster scoped resource.

### Checklist

Fixes #


JorTurFer (approved):
Nice catch! Thank you!

Issue #505, opened 2023-08-07
KEDA crashes when using cert-manager certificates and restricted secret access


zerodayyy:
When enabling restricted secret access feature and using cert-manager instead of self-managed certificates, KEDA operator and metrics-server fail to start.

A possible workaround for this is providing access to own-namespace secrets by creating Role and RoleBinding manually or via `extraObjects` in chart values.

## Expected Behavior
KEDA chart renders necessary Role and RoleBinding under these conditions
**OR**
KEDA doesn't try to access any secrets

## Actual Behavior
KEDA chart doesn't render Role and RoleBinding, leaving no access to secrets at all (which would otherwise be available through ClusterRole). KEDA components try to access secrets in own namespace, failing to do so due to missing access permissions, and crash.

## Steps to Reproduce the Problem
  1. Restrict global secret access via following Helm values:
```yaml
permissions:
  metricServer:
    restrict:
      secret: true
  operator:
    restrict:
      secret: true
```
  2. Enable cert-manager certificates instead of self-generated ones via following Helm values:
```yaml
certificates:
  autoGenerated: false
  certManager:
    enabled: true
```
  3. Deploy KEDA to cluster
  4. KEDA metrics-server and operator crash with following error:
```
W0807 18:55:16.700808       1 reflector.go:424] k8s.io/client-go/informers/factory.go:150: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:keda:keda-operator" cannot list resource "secrets" in API group "" in the namespace "keda"
E0807 18:55:16.700852       1 reflector.go:140] k8s.io/client-go/informers/factory.go:150: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:keda:keda-operator" cannot list resource "secrets" in API group "" in the namespace "keda"
W0807 18:55:18.950561       1 reflector.go:424] k8s.io/client-go/informers/factory.go:150: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:keda:keda-operator" cannot list resource "secrets" in API group "" in the namespace "keda"
E0807 18:55:18.950596       1 reflector.go:140] k8s.io/client-go/informers/factory.go:150: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:keda:keda-operator" cannot list resource "secrets" in API group "" in the namespace "keda"
[...]

PR #518, opened 2023-09-08
Fix KEDA crashes when using cert-manager certificates and restricted secret access

gjacquet:
Allow KEDA operator to get, list and watch secrets in its own namespace when restricted mode and certmanager are enabled.

### Checklist

Fixes  https://github.com/kedacore/charts/issues/505


keda/Chart.yaml
```diff
@@ -8,7 +8,7 @@ kubeVersion: ">=v1.23.0-0"
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.11.2
+version: 2.11.3
```
zroubalik, line 11: please don't update chart version, we do it on a release

keda/templates/manager/role.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.certificates.autoGenerated ( not .Values.certificates.certManager.enabled ) }}
```
JorTurFer, line 1: Maybe we should check if the restrict access is set instead of adding the role always. I mean, if this manifest has to be deployed if we are using self generated certs without cert manager OR if we set restricted access. But if I set cert manager and I don't set restricted access, this role isn't required as it's covered by the ClusterRole

keda/templates/manager/role.yaml
```diff
@@ -17,11 +17,13 @@ rules:
   resources:
   - secrets
   verbs:
+  {{- if not .Values.certificates.certManager.enabled }}
```
JorTurFer, line 20: I guess that with my previous comment, the condition here should be if self generated AND NOT cert manager

keda/templates/manager/rolebinding.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.certificates.autoGenerated ( not .Values.certificates.certManager.enabled ) }}
+{{- if or .Values.certificates.autoGenerated .Values.certificates.certManager.enabled }}
```
JorTurFer, line 1: same as the other file comment for the condition

JorTurFer (commented):
Thanks for the fix!
I think that we be more fine to not deploy the Role if it's needed (it's not a pain, but if we don't deploy not required Roles is always better). WDYT?

JorTurFer:
Hi @gjacquet 
We aim to cut a release this week, could you address the comments?

gjacquet:
@JorTurFer this should be good now.

keda/templates/manager/role.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.certificates.autoGenerated ( not .Values.certificates.certManager.enabled ) }}
+{{- if or (and .Values.certificates.autoGenerated (not .Values.certificates.certManager.enabled)) (and .Values.permissions.operator.restrict.secret .Values.certificates.certManager.enabled) }}
```
JorTurFer, line 1: If `Values.permissions.operator.restrict.secret` we need this cluster role in any case.
```suggestion
{{- if or (and .Values.certificates.autoGenerated (not .Values.certificates.certManager.enabled)) (.Values.permissions.operator.restrict.secret) }}
```

keda/templates/manager/role.yaml
```diff
@@ -17,11 +17,13 @@ rules:
   resources:
   - secrets
   verbs:
+  {{- if not (and .Values.permissions.operator.restrict.secret .Values.certificates.certManager.enabled) }}
```
JorTurFer, line 20: I don't get this, we will add "write" permissions if restrict and certManager are false. So if we se certManager false and restrict true, we won't add writing permissions and the cert-controller will fail generating the secrets.
For adding this group of actions, the restrict flag doesn't matter, as it's for cert-controller and them aren't covered by ClusterRole

```suggestion
  {{- if and .Values.certificates. autoGenerated (not .Values.certificates.certManager.enabled) }}
```

gjacquet:
applied suggestions

JorTurFer (approved):
❤️

PR #520, opened 2023-09-11
fix: explicitly set insecureSkipTLSVerify as disabled for upgrades

JorTurFer:
After adding cert-management and removing `insecureSkipTLSVerify`, it seems that during migration from the newer versions, helm doesn't remove the old value (probably because the apiservice doesn't set it explicitly).

To prevent this error, this PR explicitly sets the `insecureSkipTLSVerify: false`. This property can be removed in future versions once enough time has passed since we changed the value.

### Checklist

### DIFF
--- a/keda/templates/metrics-server/apiservice.yaml
+++ b/keda/templates/metrics-server/apiservice.yaml
@@ -27,3 +27,4 @@ spec:
   version: v1beta1
   groupPriorityMinimum: 100
   versionPriority: 100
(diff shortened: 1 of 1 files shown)


zroubalik (approved):

sidewinder12s:
Is there a process to release this into the pre-2.12 charts?

Lack of this patch appears to be the root cause of this issue: https://github.com/kedacore/keda/issues/4732

PR #531, opened 2023-09-27
feat: configurable K8s cluster domain for Cert Manager

zroubalik:
Relates: https://github.com/kedacore/keda/issues/4861


JorTurFer (approved):

JorTurFer:
The chart CI uses `main` tag, and `main` tag still doesn't support the new parameter. Once it's merged the change in core and the image is generated, we can retry these tests

PR #540, opened 2023-10-01
chore: Improve the CI on PRs to be more efficient

JorTurFer:
### Checklist

### DIFF
--- a/.github/workflows/ci-external-scaler-azure-cosmos-db.yml
+++ b/.github/workflows/ci-external-scaler-azure-cosmos-db.yml
@@ -39,26 +39,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetesVersion: [v1.24, v1.23, v1.22, v1.21, v1.20, v1.19, v1.18, v1.17]
+        kubernetesVersion: [v1.24, v1.19]
         include:
           # Images are defined on every Kind release
           # See https://github.com/kubernetes-sigs/kind/releases
         - kubernetesVersion: v1.24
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-external-scaler-azure-cosmos-db.yml
```diff
@@ -35,26 +39,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetesVersion: [v1.24, v1.23, v1.22, v1.21, v1.20, v1.19, v1.18, v1.17]
+        kubernetesVersion: [v1.24, v1.19]
```
zroubalik, line 42: Should we test here the latest version only?

@tomkerkhove

tomkerkhove: Let's follow our support matrix for KEDA and not x Kubernets versions. We should only test that, the rest is on their own risk imo

JorTurFer: So, can I merge this PR?

zroubalik: Shouldn't we replace 1.24 with the latest one?

JorTurFer: I prefer to not touch their versions. I mean, if their scaler doesn't work on 1.28? Now I have just pruned up their versions and that's all, nothing more has changed

zroubalik (approved):

PR #591, opened 2024-01-18
Add tlsConfig for ServiceMonitor

JorTurFer:
After breaking [this PR](https://github.com/kedacore/charts/pull/539) trying to fix the conflicts, I open this PR with the OP's changes (all the KUDOS to @guicholeo)

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] README is updated with new configuration values *(if applicable)* [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#documentation)

PR #572, opened 2023-11-22
feat: Introduce CloudEventSources CRD and adding ClusterName parameter

SpiritZhou:
Introudce CloudEventSources Crd and adding ClustetName Parameter

### Checklist
- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [X] README is updated with new configuration values *(if applicable)* [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#documentation)
- [X] A PR is opened to update KEDA core ([repo](https://github.com/kedacore/keda)) *(if applicable, ie. when deployment manifests are modified)*

Relates to https://github.com/kedacore/keda/pull/4968
Relates to https://github.com/kedacore/keda/issues/3533

PR #558, opened 2023-11-04
Add ciliumnetworkpolicies

QuentinBisson:
add cilium network policies support

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] README is updated with new configuration values *(if applicable)* [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#documentation)

PR #593, opened 2024-01-19
Release 2.13.0

wozniakjan:
Preparation for the KEDA release 2.13.0

### Checklist

see also: https://github.com/kedacore/keda/issues/5275


wozniakjan:
I am not entirely sure about these two commits, ptal @kedacore/keda-core-contributors 
### 1) CRD update - https://github.com/kedacore/charts/pull/593/commits/2d22f005fb1256fa2f60954593404e27d0b408bc 
I did following procedure
```bash
# template CRDs for the latest release
kustomize build https://github.com/kedacore/keda/config/crd?ref=v2.13.0 > all.yaml

# split the templated file
yq -s '"crd-" + .spec.names.plural + ".yaml"' all.yaml

# move the split files to the correct location
mv crd-*.yaml $dir/../keda/templates/crds/
```
and then copied back the helm templating manually to those CRDs but there seems to be a good amount of drift so I am sceptical of my process ...

### 2) removing deprecated webhook enable flag - https://github.com/kedacore/charts/pull/593/commits/8e863fad296475115155529f3404fcab5d1e0a61
the notes mentioned to remove it, just wanted to confirm this is the right time and place :)

tomkerkhove:
> I am not entirely sure about these two commits, ptal @kedacore/keda-core-contributors
> 
> ### 1) CRD update - [2d22f00](https://github.com/kedacore/charts/commit/2d22f005fb1256fa2f60954593404e27d0b408bc)
> I did following procedure
> 
> ```shell
> # template CRDs for the latest release
> kustomize build https://github.com/kedacore/keda/config/crd?ref=v2.13.0 > all.yaml
> 
> # split the templated file
> yq -s '"crd-" + .spec.names.plural + ".yaml"' all.yaml
> 
> # move the split files to the correct location
> mv crd-*.yaml ./keda/templates/crds/
> ```
> 
> and then copied back the helm templating manually to those CRDs but there seems to be a good amount of drift so I am sceptical of my process ...
> 

I honestly just copy the artifacts from KEDA's release and past them in the files, I do not use Kustomize for it

tomkerkhove:
> ### 2) removing deprecated webhook enable flag - [8e863fa](https://github.com/kedacore/charts/commit/8e863fad296475115155529f3404fcab5d1e0a61)
> the notes mentioned to remove it, just wanted to confirm this is the right time and place :)

@kedacore/keda-maintainers I'm not sure if we should do hard delete, or if we should change the default to true? I personally prefer the latter as people can still turn it off if they want, but recall we may have taken a dependency on it in 2.13

wozniakjan:
> Mind adding the change from #558 in this PR please?

I'm on it, thanks for reminding :)

> @kedacore/keda-maintainers I'm not sure if we should do hard delete, or if we should change the default to true? I personally prefer the latter as people can still turn it off if they want, but recall we may have taken a dependency on it in 2.13

I would be in favour of keeping it and given the default is already `true`, I can just easily change the deprecation/removal note to mention 2.13 instead?

wozniakjan:
> I honestly just copy the artifacts from KEDA's release and past them in the files, I do not use Kustomize for it

@tomkerkhove, I investigated the diff a bit closer and the culprit is the `yq` I used to auto-split the crds yaml, lot a whitespace changes and line breaks in a different location, I will replace my attempt for automation with manual labour to reduce the diff :)

keda/values.yaml
```diff
@@ -141,8 +141,6 @@ metricsServer:
     successThreshold: 1
 
 webhooks:
-  # -- Enable admission webhooks (this feature option will be removed in v2.12)
-  enabled: true
```
JorTurFer, line 145: I think that this comment was wrong because after talking about this, webhooks will continue being as optional feature. Could you revert related changes? (and update the packed chart)

wozniakjan: thanks! The change is reverted and chart with index freshly regenerated

JorTurFer (approved):

tomkerkhove (approved):

tomkerkhove:
I only see changes for trigger authentication, just to verify - The other CRDs did not change?

JorTurFer:
> I only see changes for trigger authentication, just to verify - The other CRDs did not change?

I don't think so, but let me double-check

JorTurFer:
Sorry for the noise, could we include this PR as part of the release? https://github.com/kedacore/charts/pull/591

tomkerkhove:
Merged, all we need to do is sync and re-package

wozniakjan:
> Merged, all we need to do is sync and re-package

I'm on it :)

wozniakjan:
> I don't think so, but let me double-check

I double checked, looks like only TA and CTA changed + CloudEvent CRDs got added as a new one in https://github.com/kedacore/charts/pull/572. These are the only diffs regarding CRDs I observed.

tomkerkhove:
@wozniakjan Did you verify that https://github.com/kedacore/charts/blob/main/keda/templates/crds/crd-cloudeventsources.yaml does not need any changes?

JorTurFer:
> @wozniakjan Did you verify that [`main`/keda/templates/crds/crd-cloudeventsources.yaml](https://github.com/kedacore/charts/blob/main/keda/templates/crds/crd-cloudeventsources.yaml?rgh-link-date=2024-01-19T11%3A42%3A50Z) does not need any changes?

I did it

PR #609, opened 2024-02-15
feat: add the option to disable api server registration for metrics-server

raphapr:
Add the `metricsServer.registerAPIService` option to disable external metrics API registration.

This is useful when you already have an API registration (e.g. [Datadog Cluster Agent](https://docs.datadoghq.com/containers/cluster_agent/)) and want to deploy KEDA without impacting the current external metrics API.

### Checklist

### DIFF
--- a/keda/README.md
+++ b/keda/README.md
@@ -67,7 +67,7 @@ their default values.
 | `certificates.certManager.duration` | string | `"8760h0m0s"` | Certificate duration |
 | `certificates.certManager.enabled` | bool | `false` | Enables Cert-manager for certificate management |
 | `certificates.certManager.generateCA` | bool | `true` | Generates a self-signed CA with Cert-manager. If generateCA is false, the secret with the CA has to be annotated with `cert-manager.io/allow-direct-injection: "true"` |
-| `certificates.certManager.issuer` | object | `{"generate":true,"group":"cert-manager.io","kind":"ClusterIssuer","name":"foo-org-ca"}` | Reference to custom Issuer. If issuer.generate is false, then issuer.group, issuer.kind and issuer.name are required |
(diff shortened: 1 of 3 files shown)


zroubalik (commented):
Metrics Server is mandatory to run KEDA, you cannot use the one that's shipped with Datadog for example. The implementation is different

raphapr:
@zroubalik I understand that is mandatory. My rationale is there are potential scenarios where deploying KEDA without registering its metrics server becomes useful for migration purposes.

For example, in my case, I manage a cluster in which `1beta1.external.metrics.k8s.io` is pointed to datadog; so deploying KEDA without deregistering the actual metrics provider is desired to facilitate the creation of the KEDA objects - I want to migrate the `datadogmetric` to `scaledobject` first. And subsequently, I can switch the API by using `metricsServer.registerAPIService`.

Does that make sense?

JorTurFer:
> Does that make sense?

Yes, it can make sense IMO.

I guess that we could accept that feature now because I'll give another try to the proxy for supporting multiple metric servers (there have been some poeple reluctant to the subdomain idea in control plane side), so probably we will have to integrate it sooner or later and it also allows using KEDA only for ScaledJobs

Of course, this is something that requires a good explanation in docs, but I'd say that we could continue

WDYT?

zroubalik (commented):
I understand your use case, it is a rare one though. I am not extremely happy with this addition, but I won't block it 😄 I let others to decide.

My only condition is that this is properly documented, with consequencies for users (if they enable this setting).

JorTurFer:
I've been thinking about this and I'm not totally sure. Even though I understand the use case, I'm not sure if we are in the moment for supporting this feature as the proxy is still an on-going idea. 
I won't block it either but it has to be really well documented with the side effects of this. @tomkerkhove ?

PR #620, opened 2024-04-04
fix: add configurable automountServiceAccountToken

stefan-caraiman:
This PR makes `automountServiceAccountToken` configurable for the deployment definitions.

### Checklist

Fixes #

- configurable `automountServiceAccountToken` option in pod definitions


JorTurFer:
Hello,
What is the goal of this change? I mean, this change suggests that the service account token is something optional, and it isn't. If you don't mount the service account tokens KEDA won't work as all the components have to interact with k8s api server.
Am I missing something?

stefan-caraiman:
@JorTurFer @tomkerkhove PTAL 👋

JorTurFer:
I think that you should expect the rule directly in the Azure Security Center rather than  "hacking" the rule. @tomkerkhove ?

stefan-caraiman:
No harm in setting explicit definitions though, at least imo since it seems more of a hack that the internal K8s admission controller does the volume mount of the tokens without the user being aware of it.

Nevertheless will leave it up to you to decide 👍

JorTurFer:
I'm not directly against this tbh, but I'd like to see other folks thoughts.
@zroubalik  @tomkerkhove ?

stefan-caraiman:
@JorTurFer i will go ahead and close this PR in favour of #625 thanks

Issue #624, opened 2024-04-08
Better RBAC


jkremser:
Currently, there is one service account shared across all 3 deployments, even though they need only a sub-set of those RBAC rules. Also it's not possible to set rbac per namespace when running KEDA together with the `WATCH_NAMESPACE` not being equal to `""`.

here are some of my thoughts:

I’d like to achieve a fine-tuned RBAC settings using the helm values as the input. So that we can change the values that drives the keda’s installation on our side in kedify agent, render the yaml template and just re-apply based on what rbac settings will be required in the dashboard.
here are some flaws that I found in the current helm chart:

1. currently, there is a 1 clusterrole and 1 clusterbinding that gives very wide rights for the service account that is then used by the KEDA operator, the same service account is then also reused by the webhook deployment (although webook clearly doesn’t need 99% of those rights)

1. the cluster role is too wide, giving the operator basically almost cluster admin, for instance there is a read only rule for anything anywhere

1. `watchNamespace` is not being used anyhow, no matter if we configure it to an empty string (denoting the cluster-wide keda mode) or a given namespace, the cluster role is applied

1. scattered rbac options across different section in `values.yaml` (`.Values.rbac` vs `.Values.permissions`). Can we afford breaking the backward compatibility here and merge it under `.Values.rbac` ?

1. [this](https://github.com/kedacore/charts/blob/31cc43fed3d3915773bbfc67e9db6e700c1937ed/keda/templates/manager/clusterrole.yaml#L32) looks like a typo to me, there is no standard resource called external , maybe it should have been [this](https://github.com/kedacore/charts/blob/31cc43fed3d3915773bbfc67e9db6e700c1937ed/keda/templates/metrics-server/clusterrole.yaml#L14-L19)?

Some ideas:

One can mix a cluster role with (normal - not a clusterrolebinding). This way, we can reuse the same role (the clusterrole) and apply it only to a namespace using the RoleBinding ) this will reduce the number of manifest the helm chart will generate:
- `1` cluster role
- `n` role binding per each namespace specified in `watchNamespace`

Restrict the access by white-listing:

- the allowed crds (that can appear as the `scaleTargetRef`
- the names of secrets that can be read by KEDA operator (check k explain `role.rules.resourceNames` unfortunately this doesn't work for `create` verb)

jkremser:
so the current helm chart assigns all these rights to all 3 deployments (keda operator, webhook and metrics adapter/server):

```
[...]

PR #625, opened 2024-04-08
More secure rbac

jkremser:
Issue #624

the changes rely on `WATCH_NAMESPACE` variable. So if it is empty, we basically install almost the same rbac as before (`ClusterRole` with `ClusterRoleBinding` so that it can watch for events in all the namespaces).

If it contains one or multiple namespaces (so it assumes [this change](https://github.com/kedacore/keda/pull/5631) is merged), it will use normal `RoleBinding` together with the cluster role so that these rules are namespaced for each listed namespace in that env var.

Note: Normal `RoleBinding` + `ClusterRole` works the same way as `RoleBinding` with normal `Role`, but this way we can reuse the same resource and lower the complexity of helm chart a bit.

It also contains a way to restrict the set of secrets, the operator should be able to look into. By default it's the same behavior as before, (well, if `WATCH_NAMESPACE` is not empty, it will be restricted only to secrets in that ns), but one can also specify: `--set permissions.operator.restrict.namesAllowList="{foo,bar}"` so that the operator can read only secrets in that namespace called `"foo"` and `"bar"`

The other RBAC-related change is the whitelisting of CRDs that can be scaled by the operator. Again, by default it's the same behavior as before, but if requested, one can restrict the set of CRDs that can be referenced by scaled object in advance during the helm installation by:

```
rbac:
  enabledCustomScaledRefKinds: true
  scaledRefKinds:
   - apiGroup: argoproj.io
     kind: Rollout
   - apiGroup: cluster.x-k8s.io
     kind: MachineDeployment
   - apiGroup: cluster.x-k8s.io
     kind: MachinePool
```

The PR also splits the RBAC for all three individual components so that webhooks or metric server don't have the same rights as the operator itself(that actually [had](https://github.com/kedacore/charts/blob/3fb8a8d7401c33b5c72810cb7d673b3ce8545665/keda/templates/manager/clusterrole.yaml#L56-L61) a right to read ("get") every single resource in the cluster). Now each component has its own service account with a taylored roles attached to it.

Since the PR is quite complex to review, I've prepared some deployment scenarios and captured the resulting RBAC using the [`rbac-tool`](https://github.com/alcideio/rbac-tool) kubectl plugin.

- Scenario 1 - no restrictions, just separate RBACs for each KEDA component (which is common to each scenario)
  <details><summary>values.yaml</summary>
  
  ```bash
    crds:
      install: true
    image:
      keda:
        repository: docker.io/jkremser/keda # this contains the unreleased change for watching multiple explicitly mentioned namespaces
        tag: latest
      webhooks:
        repository: docker.io/jkremser/keda-admission-webhooks
        tag: latest
    
    logging:
      operator:
        level: debug
        format: json
        stackTracesEnabled: true
    ```
  </details>
  
  final rbac:
    - `operator`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-status-quo-operator.html)
    - `webhook`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-status-quo-webhook.html)
    - `metrics-server`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-status-quo-metrics-server)

- Scenario 2 - some CRDs are whitelisted, watching `default` ns + 2 whitelisted secrets
  <details><summary>values.yaml</summary>
  
  ```bash
      # the values.yaml from Scenario 1 + the following:
      watchNamespace: default
      rbac:
        enabledCustomScaledRefKinds: true
        scaledRefKinds:
         - apiGroup: argoproj.io
           kind: Rollout
         - apiGroup: cluster.x-k8s.io
           kind: MachineDeployment
         - apiGroup: cluster.x-k8s.io
           kind: MachinePool
      
      permissions:
        operator:
          restrict:
            namesAllowList:
             - secretPassword
             - someApiKey

    ```
  </details>
  
  final rbac:
    - `operator`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-some-scaleable-crds-operator.html)
    - `webhook`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-some-scaleable-crds-webhook.html)
    - `metrics-server`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-some-scaleable-crds-metrics-server)

- Scenario 3 - all CRDs are allowed + some namespaces are being watched + some secrets are whitelisted
  <details><summary>values.yaml</summary>
  
  ```bash
      # the values.yaml from Scenario 1 + the following:
      rbac:
        enabledCustomScaledRefKinds: true
      watchNamespace: default,keda,production
      permissions:
        operator:
          restrict:
            namesAllowList:
             - foo
             - bar
             - baz
    ```
  </details>
  
  final rbac:
    - `operator`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-all-scaleable-crds-operator.html)
    - `webhook`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-all-scaleable-crds-webhook.html)
    - `metrics-server`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-all-scaleable-crds-metrics-server)

- Scenario 4 - all namespaces are watched, only secrets called `"bar"` can be read
  <details><summary>values.yaml</summary>
  
  ```bash
      # the values.yaml from Scenario 1 + the following:
      watchNamespace: "" # all namespaces -> cluster-wide rights
      enabledCustomScaledRefKinds: false
      permissions:
        operator:
          restrict:
            namesAllowList:
             - bar
    ```
  </details>
  
  final rbac:
    - `operator`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-all-namespaces-operator.html)
    - `webhook`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-all-namespaces-webhook.html)
    - `metrics-server`: [link to visualization](https://jkremser.github.io/keda-charts/rbac-all-namespaces-metrics-server)

note: for all four scenarios, I've created a Deployment, ScaledObject, checked that hpa was created and deleted it and was checking if there were any rbac related issues


jkremser:
todo:
```
2024-04-08T15:27:07Z	ERROR	setup	unable to set up cert rotation	{"error": "secrets is forbidden: User \"system:serviceaccount:default:keda-operator\" cannot list resource \"secrets\" in API group \"\" in the namespace \"default\""}
main.main
```

(`--set watchNamespace=default`)

JorTurFer:
WOW, this PR is a great idea! I'm checking it in depth because you said, it's complex and RBAC is risky, but I'll post something asap

keda/templates/manager/clusterrole.yaml
```diff
@@ -26,77 +30,59 @@ rules:
...
 - apiGroups:
   - '*'
   resources:
   - '*/scale'
   verbs:
-  - get
-  - list
-  - patch
-  - update
-  - watch
```
JorTurFer, line 71: I think that we can just keep the previous verbs as they blocked the deletions an * grants the removal permission too

keda/templates/manager/clusterrole.yaml
```diff
@@ -26,77 +30,59 @@ rules:
...
-  {{- if and .Values.certificates.autoGenerated ( not .Values.certificates.certManager.enabled ) }}
```
JorTurFer, line 72: `admissionregistration.k8s.io` and `apiregistration.k8s.io` permissions are required in the operator when certificates are managed internally. I've seen that you moved this permission from operator to webhooks but the operator is the correct place for it

keda/templates/manager/clusterrole.yaml
```diff
@@ -26,77 +30,59 @@ rules:
...
-  - validatingwebhookconfigurations
+  - deployments/scale
+  - statefulsets/scale
   verbs:
-  - get
-  - list
-  - patch
-  - update
-  - watch
+  - '*'
```
JorTurFer, line 67: Same as above, with this change we are getting the `delete` permission when it's not required

keda/templates/manager/clusterrole.yaml
```diff
@@ -26,77 +30,59 @@ rules:
...
+  - {{ .kind | quote }}
+  - {{ printf "%s/scale" .kind | quote }}
   verbs:
-  - get
-  - list
-  - patch
-  - update
-  - watch
-  {{- end }}
+  - '*'
```
JorTurFer, line 75: same as above

keda/templates/manager/clusterrole.yaml
```diff
@@ -26,77 +30,59 @@ rules:
...
+  {{- end }}   
+{{- end }}
+{{- end }}
 - apiGroups:
   - apps
   resources:
   - deployments
   - statefulsets
   verbs:
+  - get
```
JorTurFer, line 85: I think that watch allows `get` implicitly (although I don't think that adding it explicitly is a problem, so we can keep it)

keda/templates/manager/minimal-rbac.yaml
```diff
@@ -0,0 +1,120 @@
...
+    app.kubernetes.io/name: {{ .Values.operator.name }}-minimal-cluster-role
+    {{- include "keda.labels" . | indent 4 }}
+  name: {{ .Values.operator.name }}-minimal-cluster-role
+rules:
+- apiGroups:
+  - keda.sh
+  resources:
+  - clustertriggerauthentications
+  verbs:
+  - '*'
```
JorTurFer, line 77: I think that this permissions is already granted in the operator rbac

JorTurFer (commented):
Looking nice! @zroubalik  @tomkerkhove , we should test this really well as it's an important change related with RBAC. 
After merging this, we will have to be careful when we apply RBAC changes from core to the chart

keda/templates/metrics-server/clusterrolebinding.yaml
```diff
@@ -57,6 +59,6 @@ roleRef:
   name: {{ .Values.operator.name }}-external-metrics-reader
 subjects:
 - kind: ServiceAccount
-  name: horizontal-pod-autoscaler
-  namespace: kube-system
```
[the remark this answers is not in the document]
jkremser: to be honest I wasn't sure about this one. I did a brief research by checking who is this service account and I didn't find the service account called `horizontal-pod-autoscaler` in `kube-system` ns in any k8s cluster I have running so I thought it's a obsolete thing. However, I found it now on kind and k3d. It wasn't present in GCP and AWS that used the "virtualized control planes".

To my knowledge hpa controller runs within the kube-controller-manager which doesn't use such service account. But I learned that there is a flag for `kube-controller-manager` process called `--use-service-account-credentials` that acutally splits the rbac for individual controllers contained in the `kube-controller-manager`

there is also that clusterrole that has
```
k describe clusterrole system:kube-controller-manager
  *.*                                        []                 []                         [list watch]
```

that might capture this, but just to be sure, I am adding this back.

keda/templates/metrics-server/clusterrolebinding.yaml
```diff
@@ -37,9 +38,10 @@ roleRef:
   name: extension-apiserver-authentication-reader
 subjects:
 - kind: ServiceAccount
-  name: {{ .Values.serviceAccount.name }}
+  name: {{ (.Values.serviceAccount.metricServer).name | default .Values.serviceAccount.name }}
   namespace: {{ .Release.Namespace }}
 ---
+# todo this
```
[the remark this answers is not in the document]
jkremser: Currently, there is just one service account for all 3 deployments that's configurable by `.Values.serviceAccount.*`. I am splitting this to individual 3 service accounts that can have their own settings in the values.yaml. 
However, to stay backward compatible, I guess we should allow users to use their old values.yaml w/ new helm chart version, so I choose this "fallback" approach. Where it first tries the new property and if it's not there (user didn't updated their values.yaml yet) it fall backs to the old path and at the same time he/she will be notified in the `NOTES.txt` that they are using a deprecated api.

I agree, It's kind of ugly. I was thinking having it there for some transition period, where the deprecation is in place and then remove the fallback parts (`(.Values.serviceAccount.metricServer).name | default <this>`)

keda/templates/manager/minimal-rbac.yaml
jkremser: you are right, but I'd leave it here and remove it from the other place. Because these 2 `(Cluster)Role+(Cluster)RoleBinding` are the bare minimum that is needed for operator to even start. The `clustertriggerauthentications` is a cluster-wide CRD so it doesn't make sense to include it in the "`watch_namespace` logic".

keda/templates/manager/clusterrole.yaml
jkremser: you are right, I didn't realise this. I am adding it back so that deletes are not possible. Good catch.

keda/templates/manager/clusterrole.yaml
jkremser: for operator, it's still there in the `minimal-rbac.yaml` (because it's a cluster-wide resource). I wasn't sure if this is needed by webhook itself or not so I left it there just for sure, but I will remove it. I've double checked and operator needs this for cert rotation and when installing the webhook the `ValidatingWebhookConfiguration` is created as a normal resource by helm and applied by helm (so probably someone with superpowers rbac :D)

keda/templates/webhooks/clusterrole.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - apiregistration.k8s.io
+  resources:
+  - apiservices
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+{{- end }}
```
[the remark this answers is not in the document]
jkremser: removing, good catch

keda/templates/metrics-server/clusterrolebinding.yaml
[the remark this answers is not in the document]
jkremser: ah, I missed the `todo comment`, right I put it there, because I wasn't sure about this one. But let's rather keep it there, because on Kind and k3d the service account actually really exist and on other k8s distros it won't hurt, the rollbinding can be created even if the SA is not there. So I am removing the "todo" comment

keda/templates/metrics-server/clusterrolebinding.yaml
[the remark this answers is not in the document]
JorTurFer: yeah, we need to keep it because depending on the control plane config, this can be required or not and I think that it's worth adding it always as it's something quite in depth for the average (managed k8s) user: https://docs.datadoghq.com/security/default_rules/cis-kubernetes-1.5.1-1.3.3/

keda/README.md
```diff
@@ -111,11 +111,9 @@ their default values.
 | `priorityClassName` | string | `""` | priorityClassName for all KEDA components |
 | `rbac.aggregateToDefaultRoles` | bool | `false` | Specifies whether RBAC for CRDs should be [aggregated](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) to default roles (view, edit, admin) |
 | `rbac.create` | bool | `true` | Specifies whether RBAC should be used |
+| `rbac.enabledCustomScaledRefKinds` | bool | `true` | Whether rbac for unseen crds that can have a scale subresource should be created |
```
wozniakjan, line 114: nit: to better match rest of the docs
```suggestion
| `rbac.enabledCustomScaledRefKinds` | bool | `true` | Whether RBAC for configured CRDs that can have a `scale` subresource should be created |
```

keda/README.md
```diff
@@ -111,11 +111,9 @@ their default values.
...
+| `rbac.scaledRefKinds` | string | `nil` | List of custom resources that support the scale subresource that can be referenced by `scaledobject.spec.scaleTargetRef` the feature needs to be also enabled by `enabledCustomScaledRefKinds` If left empty, rbac for `apiGroups: *` & `resources: */scale` will be created note: Deployments and StatefulSets are supported out of the box |
```
wozniakjan, line 115: nit: minor formatting tweaks
```suggestion
| `rbac.scaledRefKinds` | string | `nil` | List of custom resources that support the `scale` subresource and can be referenced by `scaledobject.spec.scaleTargetRef`. The feature needs to be also enabled by `enabledCustomScaledRefKinds`. If left empty, RBAC for `apiGroups: *` and `resources: */scale` will be created.
note: `Deployments` and `StatefulSets` are always enabled |
```

keda/values.yaml
```diff
@@ -275,16 +275,44 @@ rbac:
   # -- Specifies whether RBAC for CRDs should be [aggregated](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) to default roles (view, edit, admin)
   aggregateToDefaultRoles: false
 
+  # -- Whether rbac for unseen crds that can have a scale subresource should be created
```
wozniakjan, line 278:
```suggestion
  # -- Whether RBAC for configured CRDs that can have a `scale` subresource should be created
```

keda/values.yaml
```diff
@@ -275,16 +275,44 @@ rbac:
...
+  enabledCustomScaledRefKinds: true
+  # -- List of custom resources that support the scale subresource that can be referenced by `scaledobject.spec.scaleTargetRef`
+  # the feature needs to be also enabled by `enabledCustomScaledRefKinds`
+  # If left empty, rbac for `apiGroups: *` & `resources: */scale` will be created
+  # note: Deployments and StatefulSets are supported out of the box
```
wozniakjan, line 283:
```suggestion
  # -- List of custom resources that support the `scale` subresource and can be referenced by `scaledobject.spec.scaleTargetRef`.
  # The feature needs to be also enabled by `enabledCustomScaledRefKinds`.
  # If left empty, RBAC for `apiGroups: *` and `resources: */scale` will be created
  # note: Deployments and StatefulSets are supported out of the box
```

wozniakjan (commented):
looks really good, here are couple of minor formatting nits

keda/README.md
JorTurFer: This file is selfgenerated via helm-docs  from values.yaml, could you update it directly?

jkremser: he made (/suggested and i accepted via button) the changes to both actually. I've also double checked and re-run helm-docs locally and it didn't detect any change

jkremser: he made (well, suggested and i accepted) the change actualy to both files and I checked that locally by running the helm-docs (there was no drift)

keda/templates/NOTES.txt
```diff
@@ -59,6 +59,12 @@ WARNING - Running on unsupported Kubernetes version "1.{{.Capabilities.KubeVersi
 -------------------------------------------------------------------------------------
 {{- end }}
 
+{{- if .Values.serviceAccount.name }}
+-------------------------------------------------------------------------------------
+WARNING - .serviceAccount.name has been deprecated, please migrate to newest version of the Helm Chart values that allows overriding the service account name for each KEDA component
```
tomkerkhove, line 64: Let's list the knobs they can use in the new version

jkremser: done in https://github.com/kedacore/charts/pull/625/commits/1ad805c673c72c6325196df1408be7acd8db8b14

zroubalik (approved):

joebowbeer:
@jkremser Should the instructions for restricting access to secrets be updated?

https://keda.sh/docs/2.15/operate/cluster/#restrict-secret-access

Also see https://github.com/kedacore/keda-docs/issues/1307

PR #636, opened 2024-05-01
fix: ServiceAccount are correctly annotated

JorTurFer:
**NOTE**: This [other PR](https://github.com/kedacore/charts/pull/637) is in conflict as both updates the index, we will have to merge one and the rebase the other one

During the RBAC modifications, we introduced a but that suppressed the operator service account annotations. This can be critical for users who don't use `podIdentity` section but directly annotate the service account with pod identity information (basically, KEDA doesn't use pod identity on their cases)

This PR keeps the serviceaccount segregation, but instead of a fallback condition that includes the old `serviceAccount.annotations` only if the more new ones aren't set (which was failing because the values.yaml sets them) now we will place both together. When users migrate their yamls, they just need to add the new ones and remove the old ones, but as default, they won't have any breaking change

### Checklist

Fixes https://github.com/kedacore/keda/issues/5758


jkremser:
can you please elaborate on the issue here?

> only if the more new ones aren't set (which was failing because the values.yaml sets them)

what do you mean by this exactly? User's `values.yaml` sets them? Or the default values.yaml that are bundled with the new version of the chart? Because if it was the latter, the issue would make sense to me, but these are empty:
- https://github.com/kedacore/charts/blob/v2.14.1/keda/values.yaml#L299
- https://github.com/kedacore/charts/blob/v2.14.1/keda/values.yaml#L308
- https://github.com/kedacore/charts/blob/v2.14.1/keda/values.yaml#L317

so if not provided, the old format is tried

JorTurFer:
Default values provided by chart's values.yaml set them to true:
https://github.com/kedacore/charts/blob/f9f23e91d5b03059018f7eb8d0cc34112f52760a/keda/values.yaml#L290-L317

JorTurFer:
This condition is failing because of that:
```
 {{- toYaml (.Values.serviceAccount.operator).annotations | default .Values.serviceAccount.annotations | nindent 4}}
```

As `.Values.serviceAccount.operator` is true, it takes the annotations although are empty, not using ever the `default` vale, breaking the old behaviour

jkremser:
thanks again @JorTurFer for taking this one
(lgtm)

README.md
```diff
@@ -21,8 +21,8 @@ $ helm repo add kedacore https://kedacore.github.io/charts
 $ helm search repo kedacore
 NAME                                            CHART VERSION   APP VERSION     DESCRIPTION
 kedacore/external-scaler-azure-cosmos-db        0.1.0           0.1.0           Event-based autoscaler for Azure Cosmos DB chan...
-kedacore/keda                                   2.14.1          2.14.0          Event-based autoscaler for workloads on Kubernetes
-kedacore/keda-add-ons-http                      0.7.0           0.7.0           Event-based autoscaler for HTTP workloads on Ku...
+kedacore/keda                                   2.14.2          2.14.0          Event-based autoscaler for workloads on Kubernetes
+kedacore/keda-add-ons-http                      0.8.0           0.8.0           Event-based autoscaler for HTTP workloads on Ku...
```
JorTurFer, line 25: I forgot this change during the add-on release, it's not related with this directly but it doesn't hurt if we upgrade it here either xD

PR #662, opened 2024-07-14
Add a pdb template for the http-add-on interceptor

### LINKED ISSUE #658: Missing PDB for http addon interceptors
There is no pod disruption budget on the http addon interceptors, which means that during node rollover, there might not be any interceptors available to add requests to its queue. Since the queues themselves are in memory, that means that there would be dropped requests.

Could we add a PDB to that replicaset?

abatilo:
There is no pod disruption budget on the http addon interceptors, which
means that during node rollover, there might not be any interceptors
available to add requests to its queue. Since the queues themselves are
in memory, that means that there would be dropped requests.

Closes #658

### DIFF
--- a/http-add-on/templates/interceptor/poddisruptionbudget.yaml
+++ b/http-add-on/templates/interceptor/poddisruptionbudget.yaml
@@ -1,0 +1,21 @@
+{{- if .Values.interceptor.pdb.enabled -}}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  labels:
+    app.kubernetes.io/component: interceptor
+    {{- include "keda-http-add-on.labels" . | indent 4 }}
+  name: {{ .Chart.Name }}-interceptor
+  namespace: {{ .Release.Namespace }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: interceptor
+      {{- include "keda-http-add-on.matchLabels" . | indent 6 }}
+  {{- if .Values.interceptor.pdb.minAvailable }}
+  minAvailable: {{ .Values.interceptor.pdb.minAvailable }}
+  {{- end }}
+  {{- if .Values.interceptor.pdb.maxUnavailable }}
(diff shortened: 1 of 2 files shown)


http-add-on/values.yaml
```diff
@@ -193,6 +193,15 @@ interceptor:
     # -- Port that the interceptor proxy TLS server should be started on
     port: 8443
 
+  # configuration of pdb for the interceptor
+  pdb:
+    # -- Whether to install the `PodDisruptionBudget` for the interceptor
+    enabled: false
```
tomkerkhove, line 199: Any reasons why we would not enable this by default?

http-add-on/templates/interceptor/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.interceptor.pdb.enabled -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+{{- else }}
+apiVersion: policy/v1beta1
+{{- end }}
```
tomkerkhove, line 6: > policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25.

I think it's fair to say we should not support it even

http-add-on/values.yaml
abatilo: I wasn't sure whether or not we were okay with a change that would modify the default installation. I can default to enabled!

http-add-on/templates/interceptor/poddisruptionbudget.yaml
abatilo: Great! I'll remove

abatilo:
Changes made. Please take a look when you can @tomkerkhove

tomkerkhove (approved):
LGTM, any concerns @zroubalik?

zroubalik (approved):

JorTurFer (approved):

PR #663, opened 2024-07-16
Add missing quotes for interceptor's TLS-related env vars

jkremser:
The env vars were not quoted which ended up in ports not being quoted and helm chart was not installable.

```
 {{ .. | quote }}
```
.. would also work, but this way it's consistent w/ the other variables in this file

### Checklist

Reproduction of the issue, this pr fixes:

```
helm upgrade -i http-add-on . --namespace keda  --set crds.install=false --set interceptor.tls.enabled=true
Release "http-add-on" does not exist. Installing it now.
Error: 1 error occurred:
	* Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal number into Go struct field EnvVar.spec.template.spec.containers.env.value of type string
```


zroubalik (approved):
LGTM, thanks!

PR #664, opened 2024-07-24
Add `portName` to Operator and Webhooks for Consistency

ramasai1:
The metricServer prometheus section in the KEDA chart provides a `portName` field which is the name of the port in both the deployment and the service monitor. This is not the case for the operator and admission webhooks. There's not much consistency in how the port name for the deployment is chosen vs how the service/pod monitor objects use said port.

This PR unifies the interface by allowing each component to define a `portName` and makes sure that this is the name of the HTTP port being exposed in the deployment + this is the port being used by the service/service monitor/pod monitor.

### Checklist


ramasai1:
@JorTurFer could I please get a review here?

zroubalik (commented):
We shouldn't introduce breaking changes, can we support both approaches?

PR #667, opened 2024-07-28
respect service.type

### LINKED ISSUE #648: User configure Service.type does not take effect
We have noticed an instance of incorrect helming in one of your Kubernetes configuration files. Hardcoded or default configuration values in the template are considered anti-pattern and configurable values in the values.yaml file are not being respect, resulting in user-provided configurations never being applied. Additionally, we provide anecdotal evidence from [trivy-operator#729](https://github.com/aquasecurity/trivy-operator/issues/729) regarding the incorrect helming defect.

 [Service.type](https://github.com/kedacore/charts/blob/43b0f516e0d101c18830d06126700916c029712b/keda/values.yaml#L464 ) does not take effect when changed to a value other than default value 'ClusterIP'. This is due to missing implementation for this value in metric server service.yaml file: https://github.com/kedacore/charts/blob/43b0f516e0d101c18830d06126700916c029712b/keda/templates/metrics-server/service.yaml#L28.

## Expected Behavior
Change Service.type to other type other than 'ClusterIP' can take effect

## Actual Behavior
Only default value 'ClusterIP' take effect

  1. install main version through helm chart
  2. change Service.type to any other type instead of default type 'ClusterIP'
  3. observe the service type still default type

## Specifications

  Recommend fix:
   ```yaml
   spec:
 +  type: {{ .Values.service.type }}
      ports:
      - name: https
        port: {{ .Values.service.portHttps }}
        targetPort: {{ .Values.service.portHttpsTarget }}
        protocol: TCP
   ```

   Or delete this value from documentation.

zyue110026:
Fixes #648

respect service.type.

### DIFF
--- a/keda/templates/metrics-server/service.yaml
+++ b/keda/templates/metrics-server/service.yaml
@@ -26,6 +26,7 @@ metadata:
     {{- end }}
   {{- end }}
 spec:
+  type: {{ .Values.service.type }}
   ports:
   - name: https
     port: {{ .Values.service.portHttps }}


JorTurFer:
nice catch!

JorTurFer (approved):

Issue #226, opened 2022-01-10
helm chart CRD installation first


agnewp:
## Expected Behavior
Keda can be installed from its chart as a dependency of another custom chart that uses Keda CRDs

## Actual Behavior
Helm expects that CRDs need to be installed BEFORE other resources that potentially use these CRDS.  when KEDA is a dependent chart of another one that attempts to use KEDA CRDs, the helm installation fails.

## Steps to Reproduce the Problem

  1. make a simple chart that depends on the keda chart. this simple chart should have a keda CRD in it at minimum, like a TriggerAuthentication object for example
  2. use helm to attempt installation of the top-level custom chart
  3. watch in horror as  you get this error:
 
`installation failed: unable to build kubernetes objects from release manifest: unable to recognize \"\": no matches for kind \"TriggerAuthentication\" in version \"keda.sh/v1alpha1\"`

## Specifications

  - **KEDA Version:** 1.5.1
  - **Platform & Version:**
  - **Kubernetes Version:** 1.20 eks
  - **Scaler(s):** N/A

the good news is that you aren't the first people to run into this problem, i see someone made a crude attempt at enforcing installation order by making the file names look like linux init.d scripts in some kind of assumption that helm doesn't do things in parallel.  check this out: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

the fix is supper simple; put the crd definitions into a folder called 'crds' and boom! helm will install the CRD defs first every time!

tomkerkhove:
> the fix is supper simple; put the crd definitions into a folder called 'crds' and boom! helm will install the CRD defs first every time!

This is what we used to have, but the CRDs were not being updated then. Helm does not provide a good story here.

agnewp:
well then its a trade-off. We have experienced that the current behavior for a new installation is quite flakey. the helm install command actually misses installation of some of the CRDs some of the time. also is there a point to how the files are named? does this actually work as a way to define installation order?

agnewp:
my mistake, i am using 2.5.1

tomkerkhove:
> well then its a trade-off. We have experienced that the current behavior for a new installation is quite flakey

The alternative approach will silently continue giving you incomplete CRDs.

However, I will take a look when I have time but we might want to use both approaches as a hack: `crd/` for new installations and the current approach for updates.

tomkerkhove:
Let me try the chart again tomorrow though because last time I checked I didn't have issues.

tomkerkhove:
[...]

Issue #20, opened 2020-01-31
Publish Helm chart to an OCI-compatible repo to support versioning


tomkerkhove:
When we moved to this repo, we decided to move away from hosting our charts on `https://kedacore.azureedge.net/helm` over `https://kedacore.github.io/charts`. (see https://github.com/helm/hub/pull/181)

Unfortunately, this means that we are losing our Helm chart versioning.

I'd suggest to publish our Helm chart to an OCI-compatible repo such as Azure Container Registry so that:
1. We can version our Helm charts
2. We can use releases on this repo to track changes
3. People who are using an older version can stick to an older chart

jeffhollan:
Concern with this would be that OCI support for helm repos appears to be opt-in and expiremental? https://helm.sh/docs/topics/registries/#enabling-oci-support

jeffhollan:
But like the idea of having versioned outside of just git commits.  Also do love how easy hosting a chart from a GitHub pages is (what we are doing today).  Open to options though

jeffhollan:
Actually looks like we are snapshotting versions of the chart for each release.  We currently have 2 chart versions: 1.0.0 and 1.1.0 and soon will have a 1.2.0 

https://github.com/kedacore/charts/tree/master/docs

tomkerkhove:
It looks like you are right:
```
❯ helm search repo kedacore/keda --versions
NAME            CHART VERSION   APP VERSION     DESCRIPTION
kedacore/keda   1.1.0           1.1.0           Event based autoscaler for Azure Functions depl...
kedacore/keda   1.0.0           1.0.0           Event based autoscaler for Azure Functions depl...
```

I've missed that part, sorry :/

I'd suggest to close it and keep as is then, ok?

MShekow:
Helm v3 now fully supports OCI-based Helm charts. I was wondering whether this ticket could be reopened, and whether the KEDA Helm charts could (_additionally_) be pushed to some OCI registry. 

Use case: OCI-based Helm chart hosting greatly simplifies accessing a Helm chart from a CI/CD runner when there are corporate network restrictions, with OCI-registry mirrors (which do not properly support synchronizing upstream ChartMuseum-based charts).

tomkerkhove:
I'm happy to re-evaluate this - Any guidance or want to contribute this?

MShekow:
[...]

PR #675, opened 2024-08-05
Add keda-crd helm chart

ecerulm:

### Checklist

Fixes kedacore/keda#5575


JorTurFer:
WDYT @tomkerkhove @zroubalik ? Maybe we can split the CRDs into a new fully isolated chart and deploy that chart as dependency of KEDA (managed by a flag). This could allow installing the CRDs in advantage, without losing the helm support for the chart.
This PR adds a symlink, but maybe we could iterate this

tomkerkhove:
As per https://github.com/kedacore/charts/issues/226, dependency does not work properly

JorTurFer:
No no, I meant the CRDs chart as KEDA chart dependecy, not KEDA as user dependency

tomkerkhove:
I know, but I believe the problem will be the same - There is no guarantee that CRDs are created first nor that they will be updated (biggest issue)

sofuture:
This would be quite nice to have, what's between here and :shipit: ?

pragmaticivan:
Worth mentioning that Karpenter moved away from Symlinks due to some conflict on updates across major versions.

https://github.com/aws/karpenter-provider-aws/commit/a69a37718434c583ec67996ec620fbeef2dd983a

(Mention since this came from an issue referring to Karpenter)

JorTurFer:
I still think that splitting the charts could be quite easy and solves the issues from other people, WDYT @wozniakjan @zroubalik ?

zroubalik:
Maybe include this into a bigger effort with https://github.com/kedacore/charts/issues/20

ariciputi:
Whichever way you decide to choose to solve this is fine, but having a distinct chart for CRDs would be very handy for those using Helm to deploy Keda. Thanks

JorTurFer:
I think that we can go in the same direction that other projects have taken, instead of adding syslinks to replicate the CRDs in other chart, directly splitting the CRDs into another helm chart and using it as an optional subchart in KEDA's chart. This way allows us to still manage CRDs via helm, updating them when it's needed, and also allows to install them split and manage them split if users want to do in that way for any reason

PR #711, opened 2024-11-21
keda: cert-manager: rotate CA every 9 months

### LINKED ISSUE #710: Wrong Renew Before default values in cert-manager integrations leads to certificate expired
After running Keda stably for over 2 years with the cert-manager integration we found a subtle but important bug that essentially turns into a exploding bomb after 2 years :bomb:. Our Certificate authority expired before the lead certificate had a chance to be renewed

When enabled, KEDA instanciates 2 certificates as part of a custom PKI infra to be able to communicate with the control plane using tls. In particular it creates the following:
 * A Root CA:
   * Duration: 1 year (fixed)
   * RenewBefore: 1 month (fixed)
 * A Leaf CA, signed by the Root CA:
   * Duration: 1 year (configurable)
   * RenewBefore: 8 months (configurable)

What we saw after 2 years of continual usage is described in the attached picture but I will try to summarize it in words.

image

Timeline: 
  * 17/08/24 => Lead certificate gets renewed as usual procedure
    * This will be valid for a year and won't get renewed again until 4 months have passed.
  * 21/10/24 => Root certificate gets renewed
    * The new one is valid for a year while the old one is still valid for another 30 days
  * 21/11/24 => The old CA expires. 

That last event is the problematic. Technically both certs are still valid but the leaf certificate has been signed by a cert that is expired. This leads to the communication between the metrics pod and the controller to be broken. This is very dangerous as in our case it prevented our workloads to scale correctly  as the system was unable to detect the metrics correctly.

To make the matter even more complicated cert-manager is 100% unavare. As far as the controller is concerned both certs are generated, valid and signed. No matter how many times you restart the pods nothing changes. The only solution at this point is to manually renew the leaf cert, either by manually removing the underlying secret or through the cert-manager apis.

> [!CAUTION]
> CC: @jeffhollan @JorTurFer @zroubalik @wozniakjan 
> I ask for the keda mantainers to please pay attention to this urgently as I suspect many people use the default config like we do and have a ticking bomb in their systems. All that is needed for this to trigger is the following config.

```yaml
certificates:
  certManager:
    enabled: true
```

At my company we are lucky enough to deploy very aggressively and stay up to date with release so we may be one of the first once to notice and diagnose the problem. 

Thankfully the scale from 0 to 1 worked fine as it is independent but since that point the work is being picked up by the HPA it wasn't unable to continue.

## Expected Behavior

There are many lessons learn for me. In regards to keda, I think the chart should always ensure that, the `renewBefore` field of the Root CA should always be bigger than the usage duration of the leaf certificate (i.e. `duration` - `renewBefore`). This ensures that there will be at least a renewal of the leaf certificate in between a new root ca being issued and the old CA being expired.

Furthermore the Root CA should probably last longer than the leaf certificate. Looking at `nginx-ingress`, another project with a similar setup their values are 5 years for the root and 1 year for the leaf. Furthermore they allow the default renewBefore value of 2/3 of the duration. This makes them safe and robust with plenty of time to switch to the new CA.

I will also open up a discussion with the cert-manager team if it makes sense to consider a certificate as invalid if the ca is expired and maybe renew it as well.

## Actual Behavior

Our Root CA expired faster than our leaf certificate could renew itself-

  1. Deploy the chart with 
```yaml
certificates:
  certManager:
    enabled: true
```
  2. Wait for 2 years... (I know a bit of an obscure scene)
  3. The metrics pod will log thing similar to this
```
W1120 15:26:04.275111       1 logging.go:55] [core] [Channel #1 SubChannel #2]grpc: addrConn.createTransport failed to connect to {Addr: "10.0.16.17:9666", ServerName: "keda-operator.keda.svc.cluster.local:9666", }. Err: connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2024-11-20T15:26:04Z is after 2024-11-20T03:13:23Z"
```

## Specifications

  - **KEDA Version:** 2.16.0
  - **Platform & Version:** AKS + cert-manager 
  - **Kubernetes Version:** 1.30.5
  - **Scaler(s):** Rabbitmq, although it really doesn't matter as it is a general problem

wozniakjan:
The CA is self-signed for 1 year with a rotation set 1 month before its own expiration.
The cert signed by the CA is also signed for 1 year with a rotation ~10~ 8 months before expiration (so rotating every ~2~ 4 months).

That theoretically means the CA, a month before its expiration, can sign a cert for 1 year (scheduled to rotate in ~2~ 4 months). But the cert will be rejected during TLS handshake in a month + 1 day because when evaluating the chain, the signing CA has expired at that point.

This PR adds extra rotation buffer, rotating CA every ~9~ 7 months (~3~ 5 months before expiration) so when it signs certs scheduled to rotate in ~2~ 4 months, it doesn't expire meanwhile.

### Checklist

Fixes #710

### DIFF
--- a/keda/templates/cert-manager/self-ca.yaml
+++ b/keda/templates/cert-manager/self-ca.yaml
@@ -14,7 +14,7 @@ spec:
     algorithm: RSA
     size: 2048
   duration: 8760h0m0s # 1 year
-  renewBefore: 720h0m0s # 1 month
+  renewBefore: 2160h0m0s # 3 months
   issuerRef:
     name: {{ .Values.operator.name }}-selfsigned-issuer
     kind: Issuer


tete17:
Hi @wozniakjan ,

I think it is not quite right. The defaults at the moment rotate the leaf certificate 8 months before it expires which means they are changed every 4 months not 2.

Furthermore the dates by which the leaf are renewed are user configurable. To me the best solution would be as follows:
 * Stop allowing the configuration of the `renewBefore` field in all certs and use the 2/3 duration default.
 * Increase the root ca validity to 5 years
 * Add a helm check that would fail rendering if a user specified a leaf certificate duration longer than half the time of the root CA (i.e 2.5 years)

That would always ensure there is a renewal of a leaf certificate in the overlap of a root ca renewal. A slightly more complicated math applies if we also allow the configuration of the `renewBefore` field of the leaf certificate but I don't see why we should allow such configuration.

wozniakjan:
> I think it is not quite right. The defaults at the moment rotate the leaf certificate 8 months before it expires which means they are changed every 4 months not 2.

silly me, you are of course correct, 5840h is 8 months, not 10 :)

I like your proposal, given you went through the hard part of figuring out the details, would you like to send a PR as well? I'm happy to review it. Alternatively, I should be able to craft this PR better sometime next week.

tete17:
Thanks @wozniakjan 

Sure thing :)

Let me try to sneak in the MR today and at worst monday morning.

tete17:
@wozniakjan I wasn't technicall enough with helm to do the check automatically unfortunatlly
https://github.com/kedacore/charts/pull/712

wozniakjan:
superseded by https://github.com/kedacore/charts/pull/712

PR #712, opened 2024-11-22
fix(cert-manager): Ensure there is at least one leaf certificate renewal when renewing the CA

### LINKED ISSUE #710: Wrong Renew Before default values in cert-manager integrations leads to certificate expired

tete17:
The renewBefore value for the root ca was simply too low barely giving the leaf certificate any time to renew itself. This leads to the root ca expiring before the leaf certificates expires.

By removing the renewBefore values we go back to the 2/3 default and as long as the leaf certificate is only valid for half of the root it should be fine.

### Checklist

Fixes #710

### DIFF
--- a/keda/templates/cert-manager/self-ca.yaml
+++ b/keda/templates/cert-manager/self-ca.yaml
@@ -13,7 +13,8 @@ spec:
   privateKey:
     algorithm: RSA
     size: 2048
-  duration: 43800h0m0s # 5 years
+  duration: 43800h0m0s     # 5 years
+  renewBefore: 14600h0m0s  # 1.6 year, 1/3rd of the duration
   issuerRef:
     name: {{ .Values.operator.name }}-selfsigned-issuer
     kind: Issuer
--- a/keda/templates/cert-manager/self-ca.yaml
+++ b/keda/templates/cert-manager/self-ca.yaml
@@ -13,8 +13,7 @@ spec:
   privateKey:
     algorithm: RSA
     size: 2048
-  duration: 8760h0m0s # 1 year
-  renewBefore: 720h0m0s # 1 month
+  duration: 43800h0m0s # 5 years
   issuerRef:
     name: {{ .Values.operator.name }}-selfsigned-issuer
     kind: Issuer


keda/templates/cert-manager/self-ca.yaml
```diff
@@ -13,8 +13,7 @@ spec:
   privateKey:
     algorithm: RSA
     size: 2048
-  duration: 8760h0m0s # 1 year
-  renewBefore: 720h0m0s # 1 month
+  duration: 43800h0m0s # 5 years
```
wozniakjan, line 16: with the leaf cert having back `renewBefore` which means they rotate every 4 months, the CA should rotate itself at least 5 months before it expires. Given it's now going to be signed for 5 years, to be conservative we can set `renewBefore` for CA to 1 year, meaning it will rotate every 4 years.

calculated the hours for leap year 366*24
```suggestion
  duration: 43800h0m0s    # 5 years
  renewBefore: 8784h0m0s  # 1 year
```

keda/README.md
```diff
@@ -64,15 +64,14 @@ their default values.
...
 | `certificates.certManager.enabled` | bool | `false` | Enables Cert-manager for certificate management |
 | `certificates.certManager.generateCA` | bool | `true` | Generates a self-signed CA with Cert-manager. If generateCA is false, the secret with the CA has to be annotated with `cert-manager.io/allow-direct-injection: "true"` |
 | `certificates.certManager.issuer` | object | `{"generate":true,"group":"cert-manager.io","kind":"ClusterIssuer","name":"foo-org-ca"}` | Reference to custom Issuer. If issuer.generate is false, then issuer.group, issuer.kind and issuer.name are required |
 | `certificates.certManager.issuer.generate` | bool | `true` | Generates an Issuer resource with Cert-manager |
 | `certificates.certManager.issuer.group` | string | `"cert-manager.io"` | Custom Issuer group. Required when generate: false |
 | `certificates.certManager.issuer.kind` | string | `"ClusterIssuer"` | Custom Issuer kind. Required when generate: false |
 | `certificates.certManager.issuer.name` | string | `"foo-org-ca"` | Custom Issuer name. Required when generate: false |
-| `certificates.certManager.renewBefore` | string | `"5840h0m0s"` | Certificate renewal time before expiration |
```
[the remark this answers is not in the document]
tete17: I agree. I have restored the value

keda/templates/cert-manager/self-ca.yaml
tete17: Is this cutting it too short still?

All it takes is for one user to bump the duration of the leaf more than 2 months and we are back at the same problem.

Removing a config is a win win in my opinion but if you are fixed on it I am happy to change it

wozniakjan: > Is this cutting it too short still?

maybe my math is off, let me elaborate on my logic here:
* the cert is signed for a year and is renewed 8 months before expiry => rotates every 4 months
* as long as the CA will continue to be valid for at least 4 months when signing the cert, the bug should not appear
* this suggests that a year before CA's expiration, it should rotate, neaning when it signs the last cert, it will still have 1 year validity
* the certs during that year will rotate twice already and get signed by the new CA

does this sound right? or am I missing a few months somewhere again...

> All it takes is for one user to bump the duration of the leaf more than 2 months and we are back at the same problem.

unfortunately, with the options given by the chart, we can't easily prevent users from shooting themself in the foot, but at least we can make it not broken by default

> Removing a config is a win win in my opinion

from a purely technical perspective, perhaps it would be better, but opensource projects have to worry about maintaining backwards compatibilty. Removing a configuration option from a chart will have to go through [a deprecation period](https://github.com/kedacore/governance/blob/3c60f829e4660e40ace301ce316ce3a6f1783d83/DEPRECATIONS.md?plain=1#L52). That means it can still be done, but will take a few releases. Meanwhile, we can fix the defaults so no other user gets this unpleasant surprise.

wozniakjan: I just got what you meant here, keeping the leaf cert configurable but the CA instead of rotating 1 year before expiration, rotating default 2/3 of duration. I'm ok with this :)

wozniakjan (approved):

keda/templates/cert-manager/self-ca.yaml
wozniakjan: hmm, one more thing, let's make the `renewBefore` explicit, we can make it `14600h0m0s` which should be 1/3 of 5 years. If someone used this chart with argo or flux and enabled server-side apply, the `720h0m0s` would remain there I think.

tete17: Sorry i couldnt response i was busy on a work trip.

LGTM. Thanks for merging 😁

PR #715, opened 2024-12-09
add: helm values for --enable-webhook-patching flag

KrishnaSindhur:

### Checklist

Fixes #

### DIFF
--- a/keda/templates/manager/deployment.yaml
+++ b/keda/templates/manager/deployment.yaml
@@ -82,6 +82,7 @@ spec:
           - "--zap-log-level={{ .Values.logging.operator.level }}"
           - "--zap-encoder={{ .Values.logging.operator.format }}"
           - "--zap-time-encoding={{ .Values.logging.operator.timeEncoding }}"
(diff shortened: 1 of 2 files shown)


keda/templates/manager/deployment.yaml
```diff
@@ -82,6 +82,7 @@ spec:
           - "--zap-log-level={{ .Values.logging.operator.level }}"
           - "--zap-encoder={{ .Values.logging.operator.format }}"
           - "--zap-time-encoding={{ .Values.logging.operator.timeEncoding }}"
+          - "--enable-webhook-patching={{ .Values.operator.enableWebhookPatching }}"
```
JorTurFer, line 85: Maybe we can use the flag that enables the webhooks. WDYT?
https://github.com/kedacore/charts/blob/c445ec15dd4344fbd0348f407e26bcb357fedf88/keda/values.yaml#L165-L166

KrishnaSindhur: yes i missed that, we can use that itself

JorTurFer (approved):

PR #736, opened 2025-02-21
chore: enable dependabot for GH actions

zroubalik:


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: weekly
```
zroubalik, line 6:
```suggestion
    interval: monthly
```

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
```
wozniakjan, line 4: nit: not sure if this might be `.github/`?

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+  schedule:
+    interval: monthly
```
wozniakjan, line 6: nit: maybe we would be ok to have this more frequently?

wozniakjan (approved):

JorTurFer (approved):

.github/dependabot.yml
zroubalik: I configured it the same way like the core repo, let's not add too much PRs here :)

.github/dependabot.yml
zroubalik: I configured it the same way like the core repo, we can extend the cadence if needed.

PR #751, opened 2025-05-06
Remove default resources requests and limits - they should be set exp…

evilr00t:
…licitly by the operator

### Checklist

Fixes #

Removes default values for requests and limits, allow operator to define requests and limits when needed. With current implementation it is not possible to remove CPU limits as it will be picked from the default values. 

I'm happy to introduce default values if needed (wouldn't change current behaviour) and use operator values if they're set.

### DIFF
--- a/keda/templates/manager/deployment.yaml
+++ b/keda/templates/manager/deployment.yaml
@@ -191,8 +191,6 @@ spec:
           resources:
             {{- if .Values.resources.operator }}
             {{- toYaml .Values.resources.operator | nindent 12 }}
-            {{- else }}
-            {{- toYaml .Values.resources | nindent 12 }}
             {{- end }}
         {{- if .Values.operator.extraContainers }}
         {{- toYaml .Values.operator.extraContainers | nindent 8 }}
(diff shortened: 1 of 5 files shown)


keda/templates/manager/deployment.yaml
```diff
@@ -191,8 +191,6 @@ spec:
           resources:
             {{- if .Values.resources.operator }}
             {{- toYaml .Values.resources.operator | nindent 12 }}
-            {{- else }}
-            {{- toYaml .Values.resources | nindent 12 }}
```
JorTurFer, line 195: I'm not sure about removing this fallback behavior, although I see the point of the PR, removing this can be disuptive

evilr00t: I could leave the fallback but it would fallback to the empty object which seems to be pointless? IMO resources/limits should be set explicitly.

JorTurFer: My afraid is about doing breaking changes if someone is relying on default values. @kedacore/keda-helm-contributors

wozniakjan: I don't think it's a good idea to remove the default values, this might be a breaking change for some users.

wozniakjan (commented):
> With current implementation it is not possible to remove CPU limits as it will be picked from the default values.

what is preventing you from unsetting this explicitly? e.g.
```
helm template keda kedacore/keda --namespace keda --set resources.operator.limits.cpu=null
```

I'm going to close this for now unless there is another rationale for changing defaults.

PR #758, opened 2025-06-15
feat: support different node selector and toleration for operator, metric server and webhooks

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

this PR introduces a new separate values for tolerations and node selector for metric server, operator, webhook.
i do think this is a breaking change, a way to maybe make it not a breaking change is to add the node selector and toleration under each spec (while keeping the global one) instead of doing a key per component under the nodeSelector and toleration spec. 
let me know if you want me to implement it that way.  

### Checklist


JorTurFer (changes requested):
Hello
We must avoid breaking changes. Please do it in a way that avoids breaking change, for example, defining this new fields under operator.tolerations, metrics-server.tolerations, etc and use the current value as default if the other isn't defined

KyriosGN0:
> Hello We must avoid breaking changes. Please do it in a way that avoids breaking change, for example, defining this new fields under operator.tolerations, metrics-server.tolerations, etc and use the current value as default if the other isn't defined

Hi, i updated the PR accordingly

keda/templates/manager/deployment.yaml
```diff
@@ -222,8 +222,10 @@ spec:
       {{- end }}
       nodeSelector:
         kubernetes.io/os: linux
-      {{- with .Values.nodeSelector }}
-        {{- toYaml . | nindent 8 }}
+      {{- if .Values.operator.nodeSelector }}
+        {{- toYaml .Values.operator.nodeSelector | nindent 8 }}
+      {{- else if .Values.nodeSelector }}
+        {{- toYaml .Values.nodeSelector }}
```
JorTurFer, line 228: what about something like
```suggestion
      {{- with .Values.operator.nodeSelector | default .Values.nodeSelector }}
        {{- toYaml . | nindent 8 }}
```
This applies the new value and it defaults to the previous one

KyriosGN0:
hey @JorTurFer, can i get another review here? 
Thanks!

KyriosGN0:
hey @JorTurFer, can i get another review here? 
Thanks!

JorTurFer (approved):
LGTM! Thanks for the contribution! 
@wozniakjan PTAL

JorTurFer:
> hey @JorTurFer, can i get another review here? Thanks!

Sorry for the slow response btw 😿

wozniakjan:
looks like this PR broke valid configuration for setting `.Values.tolerations` https://github.com/kedacore/charts/issues/861. The `{{ if .. }}` blocks forgot to add `tolerations:` to all branches and it's rendered only in certain cases

PR #741, opened 2025-02-28
Update helm charts to support zap logging in metrics server

aliaqel-stripe:
Update metrics-server helm charts to support https://github.com/kedacore/keda/pull/6578

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [ ] README is updated with new configuration values *(if applicable)* [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#documentation)
- [x] A PR is opened to update KEDA core ([repo](https://github.com/kedacore/keda)) *(if applicable, ie. when deployment manifests are modified)*

Fixes #

PR #768, opened 2025-07-05
Update metrics apiserver to use zap logging flags

abatilo:
The KEDA metrics apiserver deprecated klogr flags in favor of zap in
commit 4ee73f064 [1]. The adapter's main.go marks --v, --logtostderr,
and --stderrthreshold as DEPRECATED [2].

KEDA's own deployment.yaml switched to --zap-log-level [3], but the
helm chart still used the deprecated flags.

This change:
- Replaces --v with --zap-log-level
- Removes --logtostderr and --stderrthreshold
- Updates values.yaml to use string log levels (info/debug/error)
  matching other KEDA components for consistency

[1] https://github.com/kedacore/keda/commit/4ee73f064
[2] https://github.com/kedacore/keda/blob/main/cmd/adapter/main.go#L245-L250
[3] https://github.com/kedacore/keda/blob/main/config/metrics-server/deployment.yaml#L60

### DIFF
--- a/keda/templates/metrics-server/deployment.yaml
+++ b/keda/templates/metrics-server/deployment.yaml
@@ -122,15 +122,16 @@ spec:
           args:
           - --port={{ .Values.prometheus.metricServer.port }}
           - --secure-port={{ .Values.service.portHttpsTarget }}
-          - --logtostderr=true
-          - --stderrthreshold={{ .Values.logging.metricServer.stderrthreshold }}
           - --disable-compression={{ .Values.metricsServer.disableCompression}}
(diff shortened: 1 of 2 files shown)


wozniakjan:
hey @abatilo, can you please sign the commit? It should be sufficient to `git commit --amend --signoff`.

abatilo:
Done @wozniakjan !

JorTurFer (commented):
IIRC, the metric server still supports the old values for backward compatibility, right? I guess that we should keep both option in helm for a while to avoid breaking changes

JorTurFer:
I think that this PR does it indeed -> https://github.com/kedacore/charts/pull/741
doesn't? PTAL @wozniakjan

PR #782, opened 2025-10-03
chore: update Kubernetes versions in CI workflows for compatibility

zroubalik:
Let's align with 2.18 compatibility

### Checklist

Relates: https://github.com/kedacore/keda/issues/6853

### DIFF
--- a/.github/workflows/ci-core.yml
+++ b/.github/workflows/ci-core.yml
@@ -42,7 +42,7 @@ jobs:
       fail-fast: false
       matrix:
         enableAzureWorkloadIdentity: [false, true]
-        kubernetesVersion: [v1.32, v1.31, v1.30, v1.23]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-core.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       fail-fast: false
       matrix:
         enableAzureWorkloadIdentity: [false, true]
-        kubernetesVersion: [v1.32, v1.31, v1.30, v1.23]
+        kubernetesVersion: [v1.33, v1.32, v1.31, v1.23]
```
wozniakjan, line 45: nit: do we still want to test against v1.23 given it's been more than 2 years since it is no longer supported?

zroubalik: Yeah, it is just a sanity check, it is a version after autoscaling v2beta2 -> v2 change.

We don't claim support with this version

PR #814, opened 2026-02-09
Fix remaining empty annotations: keys in chart templates

### LINKED ISSUE #397: Empty annotations in APIService cause diffs in GitOps tooling
When installing KEDA using Argo CD, the application never ends up being synced. It's caused by empty `annotations` in APIService `v1beta1.external.metrics.k8s.io` defined in `keda/templates/24-metrics-apiservice.yaml`.

Kubernetes API does not return null objects when queried, which causes Argo CD to consider this a valid diff and try syncing the application again.

image

This happens due to a bug in `keda/templates/24-metrics-apiservice.yaml` Helm template, where `annotations` map is populated even if no annotations are present:

```yaml
  annotations:
    {{- if .Values.certificates.certManager.enabled }}
    {{- if .Values.certificates.certManager.generateCA }}
    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Values.operator.name }}-ca
    {{- else }}
    cert-manager.io/inject-ca-from-secret: {{ .Release.Namespace }}/{{ .Values.certificates.certManager.caSecretName }}
    {{- end }}
    {{- end }}
    {{- if .Values.additionalAnnotations }}
    {{- toYaml .Values.additionalAnnotations | nindent 4 }}
    {{- end }}
```

which renders to:

```yaml
  annotations:
```

## Expected Behavior

When no annotations are provided with Helm values, APIService `v1beta1.external.metrics.k8s.io` does not contain `metadata.annotations`.

## Actual Behavior

When no annotations are provided with Helm values, APIService `v1beta1.external.metrics.k8s.io` contains an empty `metadata.annotations`.

  1. Install KEDA using Argo CD, without providing additional annotations
  2. Observe the above diff

BEvgeniyS:
Fix remaining empty `annotations:` keys in chart templates

  Follow-up to kedacore/charts#486 which fixed empty annotation blocks
  across almost everywhere, but we still have 3 more:

  - webhooks/service.yaml: the outer conditional included
    `prometheus.webhooks.enabled` without the
    `not serviceMonitor.enabled` guard that manager/service.yaml and
    metrics-server/service.yaml both have. When prometheus + serviceMonitor
    are both enabled (prometheus.io/* annotations aren't needed), the
    annotations block renders empty.

  - metrics-server/serviceaccount.yaml: `annotations:` rendered
    unconditionally, no outer guard at all.

  - webhooks/serviceaccount.yaml: same as above.

  Empty `annotations:` keys cause issues with ArgoCD's ServerSideApply —
  the apply controller claims ownership of the annotations field, and then
  any annotation injected by external controllers (e.g. GKE NEG
  controller adding `cloud.google.com/neg`) is treated as drift, keeping
  the app permanently OutOfSync.

  ## How to reproduce

  ```sh
  # webhooks service — needs prometheus + serviceMonitor both enabled
  helm template keda keda \
    --set prometheus.webhooks.enabled=true \
    --set prometheus.webhooks.serviceMonitor.enabled=true \
    -s templates/webhooks/service.yaml | grep -C1 'annotations:$'

  # serviceaccounts — empty annotations with default values
  helm template keda keda -s templates/metrics-server/serviceaccount.yaml | grep -C1 'annotations:$'
  helm template keda keda -s templates/webhooks/serviceaccount.yaml | grep -C1 'annotations:$'
   ```
### Checklist

Fixes #397 (although that particular file is already fixed, so should probably be just closed)

### DIFF
--- a/.github/workflows/ci-core.yml
+++ b/.github/workflows/ci-core.yml
@@ -42,7 +42,7 @@ jobs:
       fail-fast: false
       matrix:
         enableAzureWorkloadIdentity: [false, true]
-        kubernetesVersion: [v1.34, v1.33, v1.32, v1.23]
+        kubernetesVersion: [v1.35, v1.34, v1.33]
         namespace: ["keda", "not-keda"]
         enableCertManager: [false, true]
         include:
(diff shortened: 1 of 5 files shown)


JorTurFer (approved):
Thanks for the fix!

Issue #816, opened 2026-02-11
Collection of breaking HTTP Addon chart issues


linkvt:
This issue contains a possibly growing collection of issues with the HTTP Addon chart that we could fix with the next more major version change as it involves breaking changes.
I decided to create this single issue to group breaking changes.

- [ ] TLS field naming is inconsistent in `values.yaml` as it uses snake_case instead of camelCase, see [values.yaml](https://github.com/kedacore/charts/blob/54497ae1bf8dde2c69b1bb85c9e000f2349f7f16/http-add-on/values.yaml#L205-L217)
- [ ] [Deployment selector labels](https://github.com/kedacore/charts/blob/54497ae1bf8dde2c69b1bb85c9e000f2349f7f16/http-add-on/templates/_helpers.tpl#L10-L19) contains many fields like `.../managed-by: Helm` instead of the necessary minimum
    - as selector labels are immutable we should focus on using only required labels to allow switching tools without deleting Deployments
- [ ] Resource Limits and Requests should be verified and updated, especially after recent improvements to the interceptor
- [ ] Some env vars and values are named incorrectly, e.g. the so-called [`endpointsCachePollingIntervalMS`](https://github.com/kedacore/charts/blob/54497ae1bf8dde2c69b1bb85c9e000f2349f7f16/http-add-on/values.yaml#L173) and the [env var](https://github.com/kedacore/http-add-on/blob/25b19f777a9c8c04b042b26779da405414c4858c/interceptor/config/serving.go#L31-L32) -> align
- [ ] Introduce a standard `fullname` helper template (with `nameOverride`/`fullnameOverride` support) and use it consistently for all resource names (Deployments, ServiceAccounts, Services, RBAC, etc.), replacing raw `.Chart.Name` / `.Release.Name` usage. This aligns with Helm community conventions (e.g. ingress-nginx, cert-manager) and would also clean up the non-standard label usage (`app:`, `name:`, `keda.sh/addon:`) in favor of the [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). See #843 / #844 for context.

PR #844, opened 2026-04-24
feat(http-add-on): ClusterRole and ClusterRoleBinding use release name

### LINKED ISSUE #843: allow per namespace install  keda-add-ons-http
### Use-Case

keda-add-ons-http multiple namespaced install

- install app1 namespace with watchNamespace app1
- install app2 namespace with watchNamespace app2

### actual
when multiple install, then  clusterRole and clusterRoleBinding name is duplicated.

### expect
allow specific role, rolebinding name or add custom prefix or suffix  
or 
clusterRole and clusterRoleBinding use release name instead of chart name

paihu:
http-add-on's ClusterRole and ClusterRoleBinding use `.Release.Name` instead of `.Chart.Name`

### Checklist

Fixes #843

### DIFF
--- a/http-add-on/templates/interceptor/rbac.yml
+++ b/http-add-on/templates/interceptor/rbac.yml
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/component: interceptor
     app.kubernetes.io/name: http-add-on
     {{- include "keda-http-add-on.labels" . | indent 4 }}
-  name: {{ .Chart.Name }}-interceptor
+  name: {{ .Release.Name }}-interceptor
 rules:
 - apiGroups:
   - discovery.k8s.io
@@ -40,15 +40,15 @@ metadata:
     httpscaledobjects.http.keda.sh/version: {{ .Values.images.tag | default .Chart.AppVersion }}
     keda.sh/addon: {{ .Chart.Name }}
     app: {{ .Chart.Name }}
-    name: {{ .Chart.Name }}-interceptor
+    name: {{ .Release.Name }}-interceptor
     app.kubernetes.io/component: interceptor
     app.kubernetes.io/name: http-add-on
     {{- include "keda-http-add-on.labels" . | indent 4 }}
-  name: {{ .Chart.Name }}-interceptor
+  name: {{ .Release.Name }}-interceptor
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: {{ .Chart.Name }}-interceptor
+  name: {{ .Release.Name }}-interceptor
 subjects:
 - kind: ServiceAccount
   name: {{ .Chart.Name }}-interceptor
--- a/http-add-on/templates/operator/rbac.yml
+++ b/http-add-on/templates/operator/rbac.yml
@@ -36,11 +36,11 @@ metadata:
     httpscaledobjects.http.keda.sh/version: {{ .Values.images.tag | default .Chart.AppVersion }}
     keda.sh/addon: {{ .Chart.Name }}
     app: {{ .Chart.Name }}
-    name: {{ .Chart.Name }}-role
+    name: {{ .Release.Name }}-role
     app.kubernetes.io/component: operator
     app.kubernetes.io/name: http-add-on
     {{- include "keda-http-add-on.labels" . | indent 4 }}
-  name: {{ .Chart.Name }}-role
+  name: {{ .Release.Name }}-role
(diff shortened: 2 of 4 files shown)


linkvt:
Hi @paihu ,
thanks for the PR - I am out of office next week but will review the PR afterwards.
Notes for myself/reviewers: We have to consider whether this change is OK now (IMO yes), whether it is complete (didn't check) and if we want to improve names of other resources, possibly also just for v1 release where breakingh changes are more OK.

Fedosin:
## Review

**Context**: Issue #843 reports that installing the http-add-on chart multiple times (e.g., once per namespace with different `watchNamespace` values) fails because cluster-scoped ClusterRole/ClusterRoleBinding resources all get the same name from `.Chart.Name` (`keda-add-ons-http`), causing collisions. This PR switches those names to `.Release.Name` so each installation gets unique cluster-scoped resource names.

### What's good

- **Correct scope** -- The change is properly limited to cluster-scoped resources (ClusterRole, ClusterRoleBinding). Namespace-scoped resources (Role, RoleBinding, ServiceAccount, Deployment, Service) are left as-is, which is correct since they don't collide across namespaces.
- **Internal references are consistent** -- ClusterRoleBinding `roleRef.name` fields are updated to match the renamed ClusterRoles. ServiceAccount references in `subjects` are correctly left unchanged (SAs are namespace-scoped, and the `namespace` field in subjects disambiguates them).
- **Aligns with existing patterns** -- The `_helpers.tpl` already uses `.Release.Name` for the `app.kubernetes.io/instance` label, so this follows an established convention.

### Issues

**1. Incomplete -- aggregate ClusterRoles are missed**

`rbac-aggregateclusterroles.yaml` defines two more cluster-scoped ClusterRoles that still use `.Chart.Name`:

```yaml
# rbac-aggregateclusterroles.yaml line 6
name: {{ .Chart.Name }}-edit

# rbac-aggregateclusterroles.yaml line 30
name: {{ .Chart.Name }}-view
```

These would also collide in a multi-install scenario. They should be updated to `.Release.Name` for consistency, or an explanation provided for why they're excluded.

**2. Breaking change -- orphaned resources on upgrade**

This is a breaking change for existing installations. When a user runs `helm upgrade`, Helm tracks resources by `kind + name`. Renaming these resources means Helm will **create new** ClusterRoles/ClusterRoleBindings (with `.Release.Name`) and leave the **old ones** (with `.Chart.Name`) orphaned in the cluster. Users would need to manually clean up the stale resources.

This should be documented in the chart's upgrade notes or changelog. The PR checklist item "README is updated with new configuration values" is unchecked -- at minimum, a note about the upgrade impact and manual cleanup steps would be valuable.

**3. Minor: `name` label inconsistency in ClusterRoleBinding metadata**

In the ClusterRoleBinding labels, some labels are changed and others aren't:

```yaml
labels:
    keda.sh/addon: {{ .Chart.Name }}          # stays .Chart.Name
    app: {{ .Chart.Name }}                    # stays .Chart.Name
    name: {{ .Release.Name }}-rolebinding     # changed to .Release.Name
```

The `name` label is updated to `.Release.Name` (makes sense -- it should match the resource name), while `keda.sh/addon` and `app` stay as `.Chart.Name` (also reasonable -- they describe the chart identity). This is fine but worth a conscious acknowledgement that the labeling convention is intentional.

### Suggestions

1. Also update `rbac-aggregateclusterroles.yaml` to use `.Release.Name`.
2. Add a note in the README or CHANGELOG about the breaking change and cleanup steps for upgrades.
3. Consider linking `Fixes #843` in the PR body (it's currently empty).

paihu:
Thanks for the review.

I had initially excluded aggregateclusterroles because it differs from the other ClusterRoles in that it grants permissions to users outside the chart and can be toggled on/off via `values.yaml`, so I was treating it as a breaking change.

For the ClusterRoles I changed in this PR, my assumption was that they're only used by roles within the chart, so the change wouldn't be breaking. However, if this change is also considered breaking, then I think it makes sense to update aggregateclusterroles along with the rest.

JorTurFer (approved):

makoh1ra33:
@paihu
Thanks for addressing the review feedback and updating the aggregate ClusterRoles as well, @paihu!
Since this already has an approval from @JorTurFer, would you be able to merge this when you get a chance? 🙏

linkvt (approved):
Thanks for the reminder, LGTM for me as well and a dedicated fullname helper is noted in issue #816 .

PR #848, opened 2026-04-29
fix(crds): use crds generated by controller 0.14.0

### LINKED ISSUE #847: InterceptorRoute CRD shipped in chart 0.14.0 does not match the controller's CRD (`spec.coldStart.fallback` schema mismatch + missing `spec.timeouts`)
## Summary

The `InterceptorRoute` CRD packaged in the `keda-add-ons-http` Helm chart for app version `0.14.0` does not match the CRD generated by the controller in [`[kedacore/http-add-on@v0.14.0](https://github.com/kedacore/http-add-on/blob/v0.14.0/config/crd/bases/http.keda.sh_interceptorroutes.yaml)`](https://github.com/kedacore/http-add-on/blob/v0.14.0/config/crd/bases/http.keda.sh_interceptorroutes.yaml).

The `spec.coldStart.fallback` schema is structurally different between the two: in the chart, `service` is a plain string; in the controller CRD (and in the Go types it's generated from), `service` is a nested object with `name` / `port` / `portName`. Because of this mismatch, the controller fails to unmarshal `InterceptorRoute` resources that have been accepted by the API server using the chart's schema.

The chart's schema is also missing the `spec.timeouts` block entirely.

## Affected versions

- Chart: `keda-add-ons-http` packaged for app version `0.14.0`
- File: [`[http-add-on/templates/crd.yaml](https://github.com/kedacore/charts/blob/main/http-add-on/templates/crd.yaml)`](https://github.com/kedacore/charts/blob/main/http-add-on/templates/crd.yaml) (current `main`)

## Source of truth

The CRD is generated by `controller-gen` from the Go types in the `http-add-on` repo. The reference CRD shipped at the v0.14.0 tag is here:

- https://github.com/kedacore/http-add-on/blob/v0.14.0/config/crd/bases/http.keda.sh_interceptorroutes.yaml

Both files have the same `controller-gen.kubebuilder.io/version: v0.19.0` annotation, so they're supposed to be regenerated together — they're not.

### 1. `spec.coldStart.fallback` — structural mismatch

**Chart (`http-add-on/templates/crd.yaml`, current `main`):**

```yaml
coldStart:
  properties:
    fallback:
      properties:
        port:
          format: int32
          maximum: 65535
          minimum: 1
          type: integer
        portName:
          minLength: 1
          type: string
        service:                  # <-- plain string
          minLength: 1
          type: string
      required:
        - service
      type: object
      x-kubernetes-validations:
        - message: exactly one of 'port' or 'portName' must be set
          rule: has(self.port) != has(self.portName)
```

**Controller (`kedacore/http-add-on@v0.14.0`):**

```yaml
coldStart:
  properties:
    fallback:
      properties:
        service:                  # <-- nested object
          properties:
            name:
              minLength: 1
              type: string
            port:
              format: int32
              maximum: 65535
              minimum: 1
              type: integer
            portName:
              minLength: 1
              type: string
          required:
            - name
          type: object
          x-kubernetes-validations:
            - message: exactly one of 'port' or 'portName' must be set
              rule: has(self.port) != has(self.portName)
      required:
        - service
      type: object
```

This looks like the chart's `fallback` schema was copied from `HTTPScaledObject.spec.coldStartTimeoutFailoverRef` (which legitimately has a flat `service: <string>` shape) and then never updated to match the actual `InterceptorRoute` Go type.

### 2. `spec.timeouts` — missing in the chart

The controller CRD has a top-level `spec.timeouts` block:

```yaml
timeouts:
  description: Timeout configuration for request handling.
  properties:
    readiness:
      type: string
    request:
      type: string
    responseHeader:
      type: string
  type: object
```

This block is **absent** from the chart's CRD. Users cannot set readiness/request/responseHeader timeouts on an `InterceptorRoute` when the CRD is installed via the chart — the API server strips unknown fields.

### 3. Minor: `scalingMetric.*.targetValue` defaults & required

The chart sets `default: 100` on both `scalingMetric.concurrency.targetValue` and `scalingMetric.requestRate.targetValue`, and does not mark them as required. The controller CRD has no default and lists `targetValue` as `required`. Less critical than the two issues above, but still a divergence worth fixing in the same regeneration.

### 4. Minor: missing `additionalPrinterColumns`

The chart's CRD does not declare the `additionalPrinterColumns` (`TargetService`, `Ready`, `Age`) that the controller CRD has, so `kubectl get interceptorroute` returns a less useful output than upstream intends.

## Reproduction

Install the chart at `0.14.0`, then apply this manifest, which matches the shape the controller's Go type expects:

```yaml
apiVersion: http.keda.sh/v1beta1
kind: InterceptorRoute
metadata:
  name: demo
  namespace: default
spec:
  target:
    service: my-app
    port: 8080
  scalingMetric:
    requestRate:
      targetValue: 100
  coldStart:
    fallback:
      service:
        name: my-fallback
        port: 8080
  rules:
    - hosts: ["demo.example.com"]
```

→ The API server **rejects** it because the chart's schema declares `coldStart.fallback.service` as a string.

If you flatten it to match the chart's schema:

```yaml
  coldStart:
    fallback:
      service: my-fallback
      port: 8080
```

→ The API server **accepts** it, but the controller fails to unmarshal the resource (the JSON decoder tries to deserialize a string into a struct), so the `InterceptorRoute` never reconciles.

In other words, **no valid manifest exists** that satisfies both the CRD validation and the controller — `InterceptorRoute` is effectively unusable when installed via the chart.

## Suggested fix

Regenerate `http-add-on/templates/crd.yaml` from the controller at the `v0.14.0` tag (or whichever app version the chart is targeting) — typically by copying `config/crd/bases/http.keda.sh_interceptorroutes.yaml` from the upstream repo and re-wrapping it with the `{{ if .Values.crds.install }}` guard. The same thing should be done for `httpscaledobjects` in the same file to be safe.

It might also be worth adding a CI check that diffs the chart's CRD against the upstream-generated one for the targeted app version, so this doesn't drift again.

## Workaround for users hitting this

Until the chart is fixed, install with `--skip-crds` and apply the upstream CRD directly:

```sh
kubectl apply --server-side -f \
  https://raw.githubusercontent.com/kedacore/http-add-on/v0.14.0/config/crd/bases/http.keda.sh_interceptorroutes.yaml

helm install http-add-on kedacore/keda-add-ons-http \
  --namespace keda --create-namespace \
  --skip-crds
```

(or `--set crds.install=false` if that values key is wired up — adjust per chart values).

Thanks!

EvannDev:
<!-- Thank you for contributing!

     Read more about how you can contribute in our contribution guide:
     https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md
--> 

 ---
  Summary                                                                                                                                                                                                                                              
         
  - Update InterceptorRoute CRD to match the schema generated by the http-add-on controller v0.14.0                                                                                                                                                    
  - Restructure the fallback field: service is now a nested object (with name, port, portName) instead of flat fields at the fallback level                                                                                                            
  - Remove the default value (100) from targetValue in concurrentRequests and requestRate scaling metrics — targetValue is now required                                                                                                                
  - Add new timeouts field to the spec with three sub-fields: readiness, request, and responseHeader                                                                                                                                                   


### Checklist

Fixes https://github.com/kedacore/charts/issues/847

### DIFF
--- a/http-add-on/templates/crd.yaml
+++ b/http-add-on/templates/crd.yaml
@@ -363,31 +363,37 @@ spec:
                 properties:
                   fallback:
                     description: |-
-                      Fallback service to route to when the target is scaling from zero
-                      and the wait timeout expires.
+                      Fallback target to route to when the primary backend does not become
+                      ready within the readiness timeout.
                     properties:
-                      port:
-                        description: Port number on the Service. Mutually exclusive
-                          with portName.
-                        format: int32
-                        maximum: 65535
-                        minimum: 1
-                        type: integer
-                      portName:
-                        description: Named port on the Service. Mutually exclusive
-                          with port.
-                        minLength: 1
-                        type: string
                       service:
-                        description: Name of the Kubernetes Service.
-                        minLength: 1
-                        type: string
+                        description: Kubernetes Service to use as the fallback target.
+                        properties:
+                          name:
+                            description: Name of the Kubernetes Service.
+                            minLength: 1
+                            type: string
+                          port:
+                            description: Port number on the Service. Mutually exclusive
+                              with portName.
+                            format: int32
+                            maximum: 65535
+                            minimum: 1
+                            type: integer
+                          portName:
+                            description: Named port on the Service. Mutually exclusive
+                              with port.
+                            minLength: 1
+                            type: string
+                        required:
+                        - name
+                        type: object
+                        x-kubernetes-validations:
+                        - message: exactly one of 'port' or 'portName' must be set
+                          rule: has(self.port) != has(self.portName)
                     required:
                     - service
                     type: object
-                    x-kubernetes-validations:
-                    - message: exactly one of 'port' or 'portName' must be set
-                      rule: has(self.port) != has(self.portName)
                 required:
                 - fallback
                 type: object
@@ -461,11 +467,12 @@ spec:
                     description: Scale based on concurrent request count.
                     properties:
                       targetValue:
-                        default: 100
                         description: Target concurrent request count per replica.
                         format: int32
                         minimum: 1
                         type: integer
+                    required:
+                    - targetValue
                     type: object
                   requestRate:
                     description: Scale based on request rate.
@@ -475,7 +482,6 @@ spec:
                         description: Bucket size for rate calculation within the window.
                         type: string
                       targetValue:
-                        default: 100
                         description: Target request rate per replica.
                         format: int32
                         minimum: 1
@@ -485,6 +491,8 @@ spec:
                         description: Sliding time window over which the request rate
                           is calculated.
                         type: string
+                    required:
+                    - targetValue
                     type: object
                 type: object
                 x-kubernetes-validations:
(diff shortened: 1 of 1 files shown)


linkvt:
Hi @EvannDev , thank you for the PR, this was a huge oversight. I'll check with the other maintainers again but plan to release this fix today.

linkvt (approved):

PR #869, opened 2026-06-04
Simplifying if/else to with, including copilot instructions (nindent fix follow-up)

### LINKED ISSUE #868: Missing nindent in manager/deployment tolerations v2.20.0
Chart v2.20.0 introduces ability to specify nodeSelector and tolerations per component.

The if/else path used when using globally defined tolerations, misses an `nindent` causing invalid YAML to be rendered.

(Now one could work around this by specifying at the component level. But that's introducing more complexity in values to work around a bug. We are skipping this chart version and hoping for a quick patch release.)

See: https://github.com/kedacore/charts/compare/v2.19.0...v2.20.0

## Expected Behavior

Same values as used for v2.19.0 rendering same tolerations as v2.20.0 without error.

## Actual Behavior

```
Error: YAML parse error on entertainment-keda/charts/keda/templates/manager/deployment.yaml: error converting YAML to JSON: yaml: line 133: mapping values are not allowed in this context

Use --debug flag to render out invalid YAML
```

Rendering chart with these values:

```yaml
tolerations:
  - key: example
    effect: NoSchedule
```

## Specifications

  - **Chart Version:** *v2.20.0*

TBeijen:
Follow up of #868 

Bug already fixed in #862 , but couldn't but notice the wildly inconsistent use of `if/else` and `with`. 

While at it, implemented the simpler change, and a copilot review instruction that aims to improve consistency in the long run.

Worth noting that the `test-values.yaml` generated in [workflow](https://github.com/kedacore/charts/blob/main/.github/workflows/ci-core.yml#L88), does not include tolerations, so by the looks of it there's quite some template code not covered by CI.

Lemme know if I can help in any way in improving that.

(Note to self: Should re-check for dupes when starting, getting distracted, and resuming a day later)

### Checklist

Fixes #868

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,36 @@
+# Copilot Instructions
+
+## Code review
+
+### Helm template: consistent fallback patterns
+
+When a template value has a component-level override falling back to a global default
+(e.g. `tolerations`, `affinity`, `nodeSelector`), prefer the `with`/`default` pattern
+over `if`/`else if`:
+
+```yaml
+# Preferred — single toYaml, nindent only once
+tolerations:
+{{- with .Values.component.tolerations | default .Values.tolerations }}
+  {{- toYaml . | nindent 8 }}
+{{- end }}
+
+# Avoid — duplicated toYaml, easy to forget nindent on one branch
+{{- if .Values.component.tolerations }}
+tolerations:
+  {{- toYaml .Values.component.tolerations | nindent 8 }}
+{{- else if .Values.tolerations }}
+tolerations:
+  {{- toYaml .Values.tolerations | nindent 8 }}
+{{- end }}
+```
+
+The `if`/`else` variant requires repeating `toYaml | nindent` in every branch, and a
+missing `nindent` produces invalid YAML that only surfaces at deploy time. The
+`with`/`default` variant has a single output point, eliminating that class of bug.
+
+### Helm template: nindent on every toYaml
+
+Every `toYaml` that emits block content (lists or maps) must be piped through `nindent`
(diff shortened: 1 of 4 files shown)


keda/templates/manager/deployment.yaml
```diff
@@ -251,12 +251,9 @@ spec:
       topologySpreadConstraints:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- if .Values.operator.tolerations }}
+      {{- with .Values.operator.tolerations | default .Values.tolerations }}
```
miqm, line 251: this is not only file with tolerations, apply change to other deployments

.github/copilot-instructions.md
```diff
@@ -0,0 +1,36 @@
...
+When a template value has a component-level override falling back to a global default
+(e.g. `tolerations`, `affinity`, `nodeSelector`), prefer the `with`/`default` pattern
+over `if`/`else if`:
+
+```yaml
+# Preferred — single toYaml, nindent only once
+tolerations:
+{{- with .Values.component.tolerations | default .Values.tolerations }}
+  {{- toYaml . | nindent 8 }}
+{{- end }}
```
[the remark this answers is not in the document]
TBeijen: Lol. Claude generated markdown messing up review instructions for Copilot.
I call deliberate sabbotage... 😜
Will fix.

keda/templates/manager/deployment.yaml
TBeijen: Fair enough.
Did a search on `{{- else if .Values`: Also implemented similar simplification to `affinity` in all 3 deployments. This should be all.

wozniakjan (approved):
this is much cleaner indeed, thank you!

PR #881, opened 2026-06-22
Make the operator gRPC metrics service port configurable

### LINKED ISSUE #511: Allow Change of Operator Port
It would be helpful if it is possible to change the ports KEDA operator is available on via the values.yaml file

### Use-Case

Our use case involves deployment of KEDA to AKS in a restrictive environment, which includes policies that allow deployment to only a select number of ports (not including 9666). Deployment is made in our environment using Terraform and here we can directly update values from values.yaml. Updating the port there is not available currently and so it would be helpful for us if these values were parameterised.

### Specification

- KEDA exposed ports are configurable from values.yaml

somaz94:
This makes the operator's gRPC Metrics Service port (hardcoded 9666) configurable via a new `operator.metricsServiceTargetPort` value (default 9666). It wires the existing core `--metrics-service-bind-address` flag and updates every reference to the port: the operator args and containerPort, the operator Service, the metrics server `--metrics-service-address` connect arg, and both NetworkPolicies.

Resolves #511 (deploying KEDA into restrictive environments that only allow a fixed set of ports, e.g. AKS policy where 9666 is not permitted).

The default stays 9666, so existing installs render identically apart from the now-explicit `--metrics-service-bind-address=:9666` arg.

Validation (local):
- helm lint passes.
- Default `helm template` render is unchanged except the explicit `--metrics-service-bind-address=:9666` arg; all ports stay 9666.
- With a custom port, all six references render the custom value, no 9666 left.
- kind v1.34 install with `operator.metricsServiceTargetPort=9777`: operator binds `:9777`, metrics-server logs "Connection to KEDA Metrics Service gRPC server has been successfully established ...:9777", and the external.metrics.k8s.io APIService reports Available.

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md) — the new value defaults to the current hardcoded `9666`, so existing installs are unaffected and nothing is deprecated or removed.
- [x] README is updated with new configuration values *(if applicable)* [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#documentation) — `keda/README.md` documents `operator.metricsServiceTargetPort`.
- [x] A PR is opened to update KEDA core ([repo](https://github.com/kedacore/keda)) *(if applicable, ie. when deployment manifests are modified)* — **not applicable.** The deployment manifest is modified, but only to pass the `--metrics-service-bind-address` flag that already exists in KEDA core (`cmd/operator/main.go`, default `:9666`). No core change is required.

Fixes #511


somaz94:
@wozniakjan @linkvt friendly ping — this has been green since late June (26/26, including the full 1.33/1.34/1.35 deploy matrix) and hasn't had a look yet. I also have #880 and #882 open here; happy to close or consolidate any of them if three at once is too much. Thanks!

linkvt:
Hi @somaz94 , thanks for the PRs, I can't tell you much about the keda chart as I'm not really involved there.

The thing I noticed though is that this PR is mixing service and target port with a single helm value, IMO there should be something like `operator.metricsService.targetPort` and `.port` in case the port really needs to be changed, but that's something Jan has to confirm :)

somaz94:
Thanks for taking a look @linkvt — and you're right that one value currently drives both the Service's `port` and `targetPort`. Let me explain why that was deliberate, and then I'm happy to go either way.

The value feeds **seven** renders across five templates, and they all have to agree for the metrics path to work at all:

| Template | Render |
|---|---|
| `manager/deployment.yaml` | `--metrics-service-bind-address=:PORT` |
| `manager/deployment.yaml` | `containerPort: PORT` |
| `manager/service.yaml` | `port: PORT` |
| `manager/service.yaml` | `targetPort: PORT` |
| `metrics-server/deployment.yaml` | `--metrics-service-address=…:PORT` |
| `manager/networkpolicy.yaml` | `port: PORT` |
| `metrics-server/networkpolicy.yaml` | `port: PORT` |

Two things follow. First, [#511](https://github.com/kedacore/charts/issues/511) is specifically about the port the operator **binds** on the pod (AKS policy disallows 9666), so `targetPort` is the one that actually has to move — the Service `port` is free to change but nobody has asked for it. Second, the metrics server reaches the operator through the Service DNS name at `:PORT`, so splitting the two would mean threading the Service `port` into `--metrics-service-address` while `targetPort` stays with the bind address and the container port. That's doable, it just turns one obviously-consistent value into two that can be set to a broken combination.

So it's a real trade-off rather than an oversight: a single value can't express `port != targetPort`, but it also can't be misconfigured.

@wozniakjan — since this is a values-schema decision that's awkward to change after release, could you make the call? Either is fine by me:

- **keep** `operator.metricsServiceTargetPort` as-is, or
- **switch** to `operator.metricsService.{port,targetPort}` (defaulting `port` to `targetPort` so the common case stays a one-liner)

I'll push whichever you prefer. CI is currently green (26/26, including the 1.33/1.34/1.35 deploy matrix).

JorTurFer (approved):

somaz94:
Thanks for the approval @JorTurFer. Two things before this lands:

@wozniakjan the values-schema question from August is still open. Right now one value drives both the Service `port` and `targetPort`, which cannot express `port != targetPort` but also cannot be set to a broken combination. @linkvt suggested splitting it into `operator.metricsService.{port,targetPort}`. Either is fine by me, but it is awkward to change after a release, so I would rather you pick.

The PR also still shows as blocked with the checks green, so it may need a second approver. Same for #882 if that one is ready to go.

zroubalik (approved):

PR #885, opened 2026-07-16
feat: add support for component-specific environment variables

mindw:
Add templated support for per component `env` injection. The existing `env` injection applied to all components and cannot reference the chart's helm variables and functions.

Note: the README had manual entries added to it, the PR includes an edited version which retains the manual edits.

### Checklist

Fixes #

### DIFF
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -679,9 +679,9 @@ envComponent:
   #   value: 'ENV-VALUE'
   # -- Additional Metrics Adapter container environment variables
   metricsServer: []
-  # -- Additional KEDA admission webhook container environment variables
   # - name: ENV_NAME
   #   value: 'ENV-VALUE'
+  # -- Additional KEDA admission webhook container environment variables
   webhooks: []
   # - name: ENV_NAME
   #   value: 'ENV-VALUE'
--- a/keda/values.yaml
+++ b/keda/values.yaml
@@ -671,6 +671,21 @@ env: []
 # - name: ENV_NAME
 #   value: 'ENV-VALUE'
 
+# -- Additional environment variables that will be passed onto specific KEDA components
+envComponent:
+  # -- Additional KEDA Operator container environment variables
+  operator: []
+  # - name: ENV_NAME
+  #   value: 'ENV-VALUE'
+  # -- Additional Metrics Adapter container environment variables
+  metricsServer: []
+  # -- Additional KEDA admission webhook container environment variables
+  # - name: ENV_NAME
+  #   value: 'ENV-VALUE'
+  webhooks: []
+  # - name: ENV_NAME
+  #   value: 'ENV-VALUE'
+
 # Extra volumes and volume mounts for the deployment. Optional.
 volumes:
   keda:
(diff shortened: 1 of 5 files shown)


linkvt:
Sorry I accidentally rebased this PR instead of another one via github UI, just reverted it so that you dont have any local conflicts.

mindw:
@rickbrouwer - any change for a review? there're are no maintainers listed for `keda-charts` and you've merged the most recent PR :)

mindw:
All suggestions were applied - thanks!

mindw:
@rickbrouwer any chance looking at this again?

rickbrouwer:
Not yet. I am traveling. I will look into it later.

keda/values.yaml
```diff
@@ -671,6 +671,21 @@ env: []
...
+  #   value: 'ENV-VALUE'
+  # -- Additional Metrics Adapter container environment variables
+  metricsServer: []
+  # - name: ENV_NAME
+  #   value: 'ENV-VALUE'
+  # -- Additional KEDA admission webhook container environment variables
+  webhooks: []
+  # - name: ENV_NAME
+  #   value: 'ENV-VALUE'
+
```
wozniakjan: I realize there is no single consensus how to set component-specific values in this chart, but `envComponent.operator` would be yet another way

```yaml
env:
  - name: ENV
    value: VAL

envComponent:
  operator:
    - name: ENV2
      value: VAL
```

I think this would be structured better under component-specific values nested as `operator.env`, e.g.
```yaml
env:
  - name: ENV
    value: VAL

operator:
  env:
    - name: ENV2
      value: VAL
```
wdyt?

wozniakjan (commented):
+1 for the feature, I'd flip the order and structure it slightly differently, but won't block either way if maintainer consensus thinks otherwise

keda/values.yaml
mindw: While this chart started with the suggested layout, sometime in the past it adapted the layout seen in this PR.
Switching layout again will do more harm than good given the amount of configurations already following the current layout.
Hopefully the above makes sense.

wozniakjan: are you sure about that? there is no `xyzComponent` for `xyz` in the whole chart afaict, this PR will be the first to introduce this pattern

wozniakjan: I can see these two patterns mostly

1) per component only
```yaml
feature:
  component: {}
```
2) global + per component
```yaml
feature: {}    # applied globally
component:
  feature: {}  # applied per component
```

but no example of what this PR introduces
```yaml
feature: {}       # applied globally
featureComponent: 
  component: {}   # applied per component
```
and imho the last stutters a bit and would be better to pick 1. or 2.. Given 1. would be a breaking change in this case, I'd propose to follow 2.

mindw: I'm completely fine with 2. I'll update the PR shortly.

mindw: @wozniakjan updated. hopefully it's what we're aiming for.

wozniakjan (approved):

PR #899, opened 2026-08-19
feat(external-scaler-azure-cosmos-db): add scheduling and service account support

aramirezh-dev:
## Problem

The `external-scaler-azure-cosmos-db` chart exposes only `additionalLabels`, `image`, `port` and `resources`. Two consequences:

1. **The pod cannot be scheduled on clusters where every node is tainted.** There is no `tolerations`, `nodeSelector` or `affinity`, and no way to inject them from a parent chart — a wrapper cannot add fields to a subchart's pod spec, and passing them with `--set` is silently ignored. On our AKS clusters (all node pools tainted: spot pools plus `CriticalAddonsOnly` on the system pool) the Deployment reconciles but the pod stays `Pending` indefinitely.

2. **Azure Workload Identity cannot be used.** The chart ships no ServiceAccount and sets no `serviceAccountName`, so there is nowhere to put the `azure.workload.identity/client-id` annotation. This is despite the scaler itself supporting identity-based auth through the `endpoint` / `clientId` trigger metadata (see [`CosmosDbFactory.GetChainedCredential`](https://github.com/kedacore/external-scaler-azure-cosmos-db/blob/main/src/Scaler/Services/CosmosDbFactory.cs)), which today is only reachable via a node-assigned managed identity.

The sibling `keda` and `keda-add-ons-http` charts in this repo already expose all of these.

## Changes

- `nodeSelector`, `tolerations`, `affinity` — all default to empty and render nothing when unset.
- `serviceAccount` block with `create` / `name` / `annotations`, plus a new ServiceAccount template.
- A `external-scaler-azure-cosmos-db.serviceAccountName` helper so that `create: false` falls back to the namespace `default` service account rather than referencing one that is never created.
- Regenerated `README.md` values table (helm-docs was not available locally, so the rows were written by hand in helm-docs format and ordering — please regenerate if anything differs).

## Compatibility note

There is one behaviour change for existing installs: because `serviceAccount.create` defaults to `true`, the Deployment now runs under a new `external-scaler-azure-cosmos-db` service account instead of the namespace `default`. The scaler needs no in-cluster RBAC, so this should be inert, but I am happy to default `create` to `false` instead if you would prefer strict no-op behaviour.

## Not included

No chart version bump — CONTRIBUTING.md notes that individual PRs do not perform release steps, so I left `version: 0.1.0` for maintainers to handle.

## Testing

```
helm lint ./external-scaler-azure-cosmos-db/ --strict     # 0 chart(s) failed
```

Rendered and verified:

- `tolerations` / `nodeSelector` / `affinity` appear in the pod spec when set, and are absent when unset.
- Default (`create: true`): ServiceAccount rendered, Deployment references it, annotations pass through.
- `create: false`: no ServiceAccount rendered, Deployment references `default`.
- `serviceAccount.name` override applies to both the ServiceAccount and the Deployment reference.


aramirezh-dev:
Closing — submitted prematurely on my side. Apologies for the noise.

PR #905, opened 2026-09-16
fix(http-add-on): remove legacy app label from RBAC templates

### LINKED ISSUE #904: http-add-on: legacy app label on RBAC templates is inconsistent and conflicts with additionalLabels
Hello, I found some issue 🙇 

## Summary

RBAC templates in the `http-add-on` chart (`ClusterRole`, `ClusterRoleBinding`, `Role`, `RoleBinding`) still carry a legacy `app: {{ .Chart.Name }}` label that the rest of the chart (`Deployment`, `Service`, `ServiceAccount`) no longer uses — those already standardized on `app.kubernetes.io/component` + `app.kubernetes.io/name` + `keda-http-add-on.labels` back in #740.

- `templates/scaler/rbac.yml` (ClusterRole, ClusterRoleBinding)
- `templates/interceptor/rbac.yml` (ClusterRole, ClusterRoleBinding)
- `templates/operator/rbac.yml` (Role, ClusterRole x2, RoleBinding, ClusterRoleBinding x2)

It isn't used as a selector anywhere in the chart, so it looks like a leftover from before the `app.kubernetes.io/name` migration rather than something intentional.

## Why it matters

Because these same objects also render `additionalLabels` via `keda-http-add-on.labels`, setting `additionalLabels.app` (a documented `values.yaml` field) makes the `app` key appear twice in the same mapping:

```yaml
# values.yaml
additionalLabels:
  app: keda-http-addon
```

```
$ helm template test http-add-on -f values.yaml --show-only templates/scaler/rbac.yml
```

```yaml
kind: ClusterRoleBinding
metadata:
  labels:
    ...
    app: keda-add-ons-http   # hardcoded, from templates/scaler/rbac.yml
    ...
    app: keda-http-addon     # from additionalLabels via keda-http-add-on.labels
```

`helm template`/`helm install` render this without complaint, but it's a duplicate YAML mapping key, so strict validators reject it, e.g. `kubeconform`:

```
error unmarshalling resource: error converting YAML to JSON: yaml: unmarshal errors:
  line 17: key "app" already set in map
```

## Proposal

Drop the redundant `app: {{ .Chart.Name }}` line from the 7 RBAC template locations above, bringing them in line with the rest of the chart. This is related to the broader label cleanup tracked in #816 (non-standard `app:`/`name:`/`keda.sh/addon:` labels), but scoped narrowly to just the RBAC objects since that part doesn't need a major-version bump — it's not used as a selector and removing it only affects metadata labels on RBAC objects.

haga0531:
Removes the hardcoded `app: {{ .Chart.Name }}` label from the `http-add-on` chart's `ClusterRole`/`ClusterRoleBinding`/`Role`/`RoleBinding` templates (`scaler/rbac.yml`, `interceptor/rbac.yml`, `operator/rbac.yml`). The rest of the chart already dropped this label in favor of `app.kubernetes.io/name` (#740), it isn't used as a selector anywhere in the chart, and it collides with `additionalLabels.app`, producing a duplicate YAML mapping key that strict validators like `kubeconform` reject.

See #904 for the repro and background.

Fixes #904


linkvt:
Thank you, great finding and just in time for the [0.16.0 release](https://github.com/kedacore/http-add-on/issues/1794) !

@coderabbitai review

Let's see if the rerun helps but from what I saw it looks good. We enabled coderabbit yesterday so interested if this is an issue we need to fix in the config or so.

linkvt:
@coderabbitai full review

linkvt:
Also a heads-up, we will soon release v1 with a bunch of breaking changes as part of https://github.com/kedacore/charts/issues/816 which you also already saw or at least referenced in the issue.

Thanks again!

linkvt (approved):

PR #909, opened 2026-09-18
feat: add helm chart flag to control apiserver config registration

juljaeg:
Adds a new flag to be able to deploy the metrics server without it being actively used. This is useful in migration scenarios where one does not want to change the scaling behavior for everything in one go.

### Checklist

- [ ] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/charts/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))


JorTurFer (approved):

omkar619-dev:
Hey, @juljaeg — declaring `registerAPIService: true` in `values.yaml` rather than using `| default true` in the template is the right call. The latter can never be turned off, since Go templates treat `false` as empty and `false | default true` evaluates to `true`.

One interaction I think is still open though, which is what I'd raised on the issue. Setting `registerAPIService: false` stops the APIService being created, but the operator carries on trying to patch one. `keda/templates/manager/deployment.yaml`
only emits `--enable-apiservice-patching` when `certificates.operator.apiServicePatching` is set, and that block is commented out in `values.yaml`, so by default no flag is passed at all — and the operator's own default is `true`:

    pflag.BoolVar(&enableAPIServicePatching, "enable-apiservice-patching", true,
        "Enable patching of APIService resources. Defaults to true.")

So out of the box, disabling registration leaves the cert rotator targeting an APIService that doesn't exist.

The smallest fix I can see leaves the existing branch alone and adds one ahead of it:

    {{- if not .Values.metricsServer.registerAPIService }}
    - "--enable-apiservice-patching=false"
    {{- else if and .Values.certificates.autoGenerated ( not .Values.certificates.certManager.enabled ) .Values.certificates.operator .Values.certificates.operator.apiServicePatching }}
    - "--enable-apiservice-patching={{ .Values.certificates.operator.apiServicePatching.enabled }}"
    {{- end }}

No APIService means nothing to patch, so the flag can be derived rather than left for the user to discover. If you'd rather not couple the two sections, documenting the pairing in the README entry for `registerAPIService` would cover it too, just
less safely.

Happy to push this to your branch if you have maintainer edits enabled, or open it as a follow-up once this lands — whichever suits you.

juljaeg:
@omkar619-dev Feel free to push 👍🏼  Maintainer edits are enabled

omkar619-dev:
Thanks @juljaeg! I think the maintainer-edits flag only covers people with push access to `kedacore/charts` though, which I don't have — so rather than have you add me as a collaborator on your fork, here's the diff. It's four lines in
`keda/templates/manager/deployment.yaml`, which this PR doesn't currently touch:

```gotemplate
{{- if not .Values.metricsServer.registerAPIService }}
- "--enable-apiservice-patching=false"
{{- else if and .Values.certificates.autoGenerated ( not .Values.certificates.certManager.enabled ) .Values.certificates.operator .Values.certificates.operator.apiServicePatching }}
- "--enable-apiservice-patching={{ .Values.certificates.operator.apiServicePatching.enabled }}"
{{- end }}

juljaeg:
@omkar619-dev included your fix as well! thanks!

omkar619-dev:
The one failing job looks like a flake rather than the new commit. It fails on `Error from server (NotFound): scaledobjects.keda.sh "nginx-autoscaling" not found`, and in the same log all three KEDA deployments are 1/1 — operator, metrics-apiserver and admission-webhooks all came up. So the chart is installed fine and it's the test workload that's missing i believe.

The other 11 `Without cert-manager` variants passed on this same commit, including `not-keda` on v1.33 and v1.34. The flag this commit adds doesn't vary by Kubernetes version or namespace, so if it were wrong I'd expect all 12 to fail rather than one.

It's worth a re-run.

rickbrouwer:
Re-run started

JorTurFer (approved):

wozniakjan (approved):

